PR #54, opened 2013-06-11
Change. socket:setfd() set `fd` to invalid value. 

moteus:
This is alternative to `acceptfd` method.

Change. socket:setfd() set `fd` to invalid value. 
Change. socket:setfd return previews `fd` value.
Change. socket.tcp and socket.udp can create object based on `fd`.

Add. test setfd

``` lua
  ---- server.lua
  cli = assert(srv:accept())
  -- test if worker thread take ownership and clear `fd`
  if run_worker_thread(cli:getfd()) then cli:setfd() end
  cli:close()

  ---- worker.lua
  -- wrap raw `fd` to socket object
  local sock = socket.tcp(..., "client")
  -- do work with sock object
```

### DIFF
--- a/src/tcp.c
+++ b/src/tcp.c
@@ -171,8 +171,11 @@ static int meth_getfd(lua_State *L)
 static int meth_setfd(lua_State *L)
 {
     p_tcp tcp = (p_tcp) auxiliar_checkgroup(L, "tcp{any}", 1);
-    tcp->sock = (t_socket) luaL_checknumber(L, 2);
-    return 0;
+    t_socket fd = tcp->sock;
+    if(lua_gettop(L) == 1) tcp->sock = SOCKET_INVALID;
+    else tcp->sock = (t_socket) luaL_checknumber(L, 2);
+    lua_pushnumber(L, (int) fd);
+    return 1;
 }
 
 static int meth_dirty(lua_State *L)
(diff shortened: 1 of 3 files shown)


ewestbrook:
This change no longer lines up with master, and it's not obvious to me that it should be made to.  Can it be either resolved or closed?

diegonehab:
Not sure what this was trying to solve.

moteus:
The main idea for this PR is allow to pass socket to the separate LuaVM (e.g. in separate os thread) and create clien socket in there.
So in the main LuaVM called `server:accept` then file descriptor pass to the separate LuaVM.
In this case it is important that in the main LuaVM GS should not close this socket if it passes to separate thread.

Tieske:
@moteus I think this is a useful change. I know this has been open like forever, but would you mind rebasing it on `master` and add the documentation changes?

diegonehab:
I think that adding a string as option to these functions will prevent future extensions, if there are any. Should there be an options table instead?

src/tcp.c
```diff
@@ -171,8 +171,11 @@ static int meth_getfd(lua_State *L)
 static int meth_setfd(lua_State *L)
 {
     p_tcp tcp = (p_tcp) auxiliar_checkgroup(L, "tcp{any}", 1);
-    tcp->sock = (t_socket) luaL_checknumber(L, 2);
-    return 0;
+    t_socket fd = tcp->sock;
+    if(lua_gettop(L) == 1) tcp->sock = SOCKET_INVALID;
```
daurnimator, line 175: Rather than `gettop` I think you'd want a `lua_isnoneornil` check.

PR #66, opened 2013-06-18
support for intermediate filters in ltn12.pump.all

fab13n:
It removes the optional step argument in ltn12.pump.all, but I've never seen it used and I can't think of a realistic use case for it.

It depends on the version of ltn12.sink.chain() which accepts multiple filter parameters.


ewestbrook:
This change is currently present on master.  Does anything else need to be done here?

We're reviewing open items in preparation for a release.  If action is needed here, please add a comment.  Otherwise, this issue will be closed on or after 24-Feb-2019.

Thanks!

Issue #28, opened 2013-01-24
inet_global_getaddrinfo


moteus:
1) inet_global_getaddrinfo ignore error from getnameinfo
2) On windows (I test only on XP SP3) when serv parameter is not NULL we get error 11044
so call have to be 

``` c
ret = getnameinfo(iterator->ai_addr, (socklen_t) iterator->ai_addrlen, hbuf, NULL, 0, NI_NUMERICHOST);
```

PR #72, opened 2013-07-31
Publish API used by luasec

horazont:
Hi everyone,

This is a quite large pull request and it deserves some explanation. It is one of two pull requests (the [second one going to luasec](https://github.com/brunoos/luasec/pull/5)) which shall fix the issue of duplication of code in luasocket and luasec and the issues which arise from using a dynamically linked version of luasec (see also [Prosody bug #28](https://code.google.com/p/lxmppd/issues/detail?id=298)).

The first commit publishes the WAITFD_\* constants, which are used by luasec at some places and is part of the modifications luasec makes to the luasocket headers.

The second one is larger and possibly more disputable. It publishes a whole bunch, probably more than strictly needed, of functions using the LUASOCKET_API define. This is required to allow luasec to find the symbols if loaded into the same process.

Together with the pull request which will be issued in a minute to luasec, this could improve user experience, especially on redhat systems, where the libraries are dynamically linked, a lot.

If you have any suggestions for improvements or any kind of critique, please let me know so that we can fix that quickly.

best regards,
Jonas

### DIFF
--- a/src/buffer.h
+++ b/src/buffer.h
@@ -20,6 +20,8 @@
 #include "io.h"
 #include "timeout.h"
 
+#include "common.h"
+
 /* buffer size in bytes */
 #define BUF_SIZE 8192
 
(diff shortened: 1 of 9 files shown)


alerque:
Does anybody have any input as to the current state of affairs here?

Tieske:
this is superseded by #346

PR #79, opened 2013-09-11
Added ability to set the option `reuseport` of a tcp socket.

KateAdams:
Enables the option `reuseport` to be set on TCP servers so many of them can bind to the same address.


thefosk:
+1 - Also related to #149

PR #80, opened 2013-09-12
Move #define WAITFD_* to .h

brunoos:


winterheart:
Hello.
Any news about this feature?

ConiKost:
Any news here?

alerque:
Does anybody understand what happened here or better yet what needs to happen?

ConiKost:
@winterheart ^^

winterheart:
Well since then I've opened PR #346 with some fixes for original idea.  That implementation already ported into Gentoo Linux luasocket package and used by luasec package. So I'd recommend move conversations there.

Issue #83, opened 2013-11-07
socket.http does not work with IPv6 servers


MarkMartinec:
I'm glad that luasocket now got support for IPv6 - this was long
overdue and is the main reason why Lua was hardly useful for me
so far.

But trying with today's code cloned from github (apparently a 3.0-rc1),
using an example of a HTTP client from documentation, it fails to work
with IPv6 web servers:

```lua
local io = require("io")
local http = require("socket.http")
local ltn12 = require("ltn12")
b,e = http.request {
  url = "http://[2001:1470:ff80::80:1]/",
  sink = ltn12.sink.file(io.stdout)
}
print(b,e)
```

gives:
```
$ lua-5.1 http-socket.lua
lua-5.1: /usr/local/share/lua/5.1/socket/http.lua:116: bad argument #2 to 'connect' (number expected, got string)
stack traceback:
        [C]: in function 'request'
        http-socket.lua:4: in main chunk
        [C]: ?
```

and:

```lua
local io = require("io")
local http = require("socket.http")
local ltn12 = require("ltn12")
b,e = http.request {
  url = "http://ipv6.test-ipv6.com/",
  sink = ltn12.sink.file(io.stdout)
}
print(b,e)
```

gives:
```
$ lua-5.1 http-socket.lua
nil     No address associated with name
```

It would be really nice and useful if support for IPv6 would be extended
to all components of the luasocket module.

PR #91, opened 2014-02-11
Modify http.lua to do a host lookup first, to determine the right address family to use (previously tcp was the default).

darklajid:
The attached patch might be nonsense or just as well not the right way™ to handle the problem.
I'm submitting this for discussion.

With this patch applied I can use the examples from #83 without problems: The ipv6 requests succeed just fine. Plus, as a bonus, ipv4 requests still work... :)

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -107,8 +107,16 @@ end
 local metat = { __index = {} }
 
 function _M.open(host, port, create)
-    -- create socket with user connect function, or with default
-    local c = socket.try((create or socket.tcp)())
+    -- create socket with user connect function, or do a lookup to default to
+    -- either tcp or tcp6 instead
+    if not create then
+	    create = socket.tcp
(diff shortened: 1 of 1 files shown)


4L3Xx:
I need this fix as well.

diegonehab:
I think the correct way to fix this is changing LuaSocket a bit. It will be better in the future, but it will take a bit more work. Do you guys want to help me? The idea is to have the behavior of c = tcp() and c:connect() be agnostic to family (unless, of course, you call c:bind() before c:connect()). It should just work regardless of ipv4 or ipv6. We already have tcp6() that forces ipv6. We should create tcp4() to force ipv4. There is a global socket.connect() that already behaves like this. Shall we do it? It will make the http code work without any change.

diegonehab:
Can you check branch "agnostic" please? I just made these changes. Seems to work fine here.

PR #101, opened 2014-07-09
Freebsd unix socket support

muggenhor:
This pull req ensures it is actually possible to connect() to unix sockets on FreeBSD without having to add a dummy char to the path. The same applies for bind().


ewestbrook:
Completed in #263 and #264. Thanks!

PR #122, opened 2015-01-12
Simple patch for Lua 5.3

andrewstarks:
I put this *almost hack* in and now Lua 5.3 appears to work. Feel free to do a proper job of it. I offer this up only because I think that socket support should always be available for the latest releases of Lua. :)


siffiejoe:
There is also `LUA_COMPAT_APIINTCASTS` which you could define in the rockspec/makefile instead.

andrewstarks:
There are many workable approaches. To my way of thinking, libraries should compile without having to track down compiler flags in the dependent libraries (which may not be compiled with that option), if that library purports to support it. 

So, it would seem that either some additional `#define` magic for Lua 5.3 that is more robust than what I did, or this pull request, or what have you, should be done. No?

moteus:
> So, it would seem that either some additional #define magic for Lua 5.3 that is more robust than what I > did, or this pull request, or what have you, should be done. No?

No. You need define this for LuaSocket. 
This macro turn on defeniton of compatability macro in `lua.h`/`lauxlib.h` sach as 
`#define luaL_checkint(L,n)	((int)luaL_checkinteger(L, (n)))`

But I agree that better fix source code than play with compile options in different buld systems. (make/CMake/LuaRocks).

PR #131, opened 2015-03-02
Updated ftp and http methods for better async support

Tieske:
If wrapping the `http` and `ftp` code for async support, not being able to pass `create` to the simple string based form of the methods, severly complicates wrapping it.

A large amount of code must be copied to create async behaviour. With this change it will be reduced to just a few lines and no redundant/duplicate code. See https://github.com/Tieske/copas/blob/671891f9e567dbd679c635f06174687df9e182ed/src/copas/copas/ftp.lua for the current `ftp.lua` async wrapper code.

for the record; change is fully backward compatible, not breaking.

### DIFF
--- a/doc/ftp.html
+++ b/doc/ftp.html
@@ -95,7 +95,7 @@ the FTP module:
 <!-- ftp.get ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ -->
 
 <p class=name id=get>
-ftp.<b>get(</b>url<b>)</b><br>
+ftp.<b>get(</b>url [, create]<b>)</b><br>
 ftp.<b>get{</b><br>
 &nbsp;&nbsp;host = <i>string</i>,<br>
 &nbsp;&nbsp;sink = <i>LTN12 sink</i>,<br>
@@ -114,7 +114,10 @@ ftp.<b>get{</b><br>
 The <tt>get</tt> function has two forms. The simple form has fixed
 functionality: it downloads the contents of a URL and returns it as a
 string. The generic form allows a <em>lot</em> more control, as explained
-below.
+below. The <tt>create</tt> parameter is only used in the string form,
+and will be used as alternative for <a href=tcp.html#socket.tcp><tt>socket.tcp</tt></a>
+if it is provided. In the advanced form <tt>create</tt> has to be specified
+in the arguments table.
 </p>
 
 <p class=parameters>
@@ -180,7 +183,7 @@ end
 <!-- put ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ -->
 
 <p class=name id=put>
-ftp.<b>put(</b>url, content<b>)</b><br>
+ftp.<b>put(</b>url, content [, create]<b>)</b><br>
 ftp.<b>put{</b><br>
 &nbsp;&nbsp;host = <i>string</i>,<br>
 &nbsp;&nbsp;source = <i>LTN12 sink</i>,<br>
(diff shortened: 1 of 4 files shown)


diegonehab:
The async support was already trivial in HTTP, which honors the "create" field in the request table. I prefer not to include a  second argument "create" to the simple interface because, honestly, it is not very commonly used and it does not deserve this preference over all other options. So, if you fix the FTP module to only accept a "create" field in the table version of the methods, like the HTTP module does, I'll merge the new pull request.

Tieske:
hmmm... I don't think I understand your comment...

Both the ftp and http commands have been updated to accept a create function in the simple form, so I don't understand your remark 'like the HTTP module does'...

Regarding the commonality of the

diegonehab:
I don't like the create() in the simple form. This is because create() is a pretty advanced feature. More advanced than the sink(), for instance. So why promote it to the second argument of the simple form instead of having the sink there instead? See what I mean?

There is also the issue of compatibility. Somewhere, somebody may be invoking http.request(foo()) where foo() returns two values: the string, and some other stuff that is supposed to be ignored. This change will break the code.

Finally, why is https.lua replicating all http.lua code? And how is this new argument in the simple form solve this? Why can't https.lua simply use the table form of http.lua to implement its simple form?

As for ftp.lua having a table form that honors create(), but no simple form with a second create() argument, I am totally fine with this. I would merge that pull request.

Tieske:
both FTP and HTTP support short and advanced forms. Both Copas and LuaSec mimic the behaviour of the HTTP request call. Now if, someone uses the short form on `luasec.https.request()`, then this function has to insert its ssl wrapped `create` function in the request. Which it can't in the short form.
So; code duplication is there because `copas.http.request`, `copas.https.request` and `luasec.https.request` all need to implement their own parsing of the argument from the short form to the advanced form, to be able to insert the create function.

But maybe we can facilitate both our needs in a different way; What if we export the private functions for parsing a short form into the advanced form? Let's call it `[table] = socket.http.parseRequest(url, body)`

Such that;
```lua
socket.http.request("http://www.google.com")
````
equals
```lua
socket.http.request(socket.http.parseRequest("http://www.google.com"))
````
The same could be applied to the FTP module.

I think this is a cleaner solution.

diegonehab:
So the code that is replicated in https.lua is the 16 lines of code in http.lua that convert the simple request in to a table request?

Tieske:
No it's more than that. The parsing uses other private methods in the modules, so they have to be copied along. PR #133 circumvents that because the function to parse is exported and it can access the other privates locally.

For example; I think that the ftp module for Copas could be reduced from 80 to around 20 lines. For the http module my guess is from 50 to 20. Combined 130 -> 40, so appr. 90 lines, or 70% reduction

But #133 is the preferred solution I think (if at all...)

diegonehab:
I agree 133 would be preferrable to the previous one. But I am still trying to understand the root of the problem here. I don't see any private functions used by srequest. Are you talking about trequest? If so, this can be replaced  by a call to the public request, right?

Tieske:
have look at `ftp.lua` 

provides `put` and `get`, but to mimic their behaviour with my own `create` function I need `sput()` and `sget()` as well. They in turn rely on `parse()`, which relies on the table `config` (which contains default constants, which I really hate copying, more than code itself actually).

When `parseRequest()` (as per this PR) is exported, I no longer need to copy all of those, just provide my own `get` and `put` and be done with it. And besides that it is a nice convenience function to quickly populate a request table, and then only adapt the ones you need instead of specifying the full parameter table.

PR #134, opened 2015-03-05
Compiletime constants exposed

### LINKED ISSUE #129: expose compile time constants
Some compile time constants, such as the max UDP datagram size should be exposed on the lua side to know the details of the lib at runtime.

Tieske:
fixes #129

### DIFF
--- a/doc/tcp.html
+++ b/doc/tcp.html
@@ -612,7 +612,8 @@ Returns the underling socket descriptor or handle associated to the object.
 </p>
 
 <p class=return>
-The descriptor or handle. In case the object has been closed, the return will be -1.
+The descriptor or handle. In case the object has been closed, the return will be -1. The 
+constant <tt>socket._SOCKETINVALID</tt> is the OS value for invalid sockets.
 </p>
 
 <p class=note>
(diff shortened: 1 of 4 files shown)


alerque:
Is this fix still relevant? The related issue is closed but at least this exact fix doesn't seem to have gone in. Perhaps something comparable did, or is it still unresolved?

Tieske:
This is still relevant imho

daurnimator (approved):
LGTM

btw, I only now notice that UDP_DATAGRAMSIZE is set to 8192; not the actual max value possible (66535). Should we update the value too?

alerque (changes requested):
This PR predates GitHub's maintainer permissions system for PR updates so only you @Tieske can push either a merge from (or rebase onto) `master` to get current CI checks.

Tieske:
I need to double check the comment by @daurnimator . iirc the original reason for this was Copas, UDP support; https://github.com/lunarmodules/copas/blob/master/src/copas.lua#L26

Tieske:
This is obsolete, was already merged through another PR. The #359 PR holds the final documentation updates.

PR #145, opened 2015-08-15
Export a few internal symbols to make lua-sec compile

gares:
This patch is applied in the Debian package, could you please merge it?


diegonehab:
This is complicated. We should probably change the names of these symbols to something prefixed by "luasocket", "ls", or something else. Otherwise, when linked with other libraries, the chance for collisions is too big. In fact, I recall somebody complaining about collisions before (he was compiling LuaSocket wrong, and all symbols were exported).

Tieske:
this is superseded by #346 , closing this.

PR #157, opened 2016-01-25
Fix error in ltn12 under Lua 5.3

mpeterv:
`unpack` became `table.unpack` in Lua 5.2, not available with default compilation flags in Lua 5.3.

### DIFF
--- a/src/ltn12.lua
+++ b/src/ltn12.lua
@@ -43,7 +43,7 @@ end
 -- (thanks to Wim Couwenberg)
 function filter.chain(...)
     local arg = {...}
-    local n = select('#',...)
+    local n = base.select('#',...)
     local top, index = 1, 1
     local retry = ""
     return function(chunk)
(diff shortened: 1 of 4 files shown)


siffiejoe:
While we are at it, there's also a plain `select` that probably should be `base.select` on line 45 ...

diegonehab:
It's best to do something like 

local unpack = unpack or table.unpack

early on. 

And also to fix the base.select. 

Do you want to redo the pull request, or shall I fix it on my side?

mpeterv:
I'll fix it. I can also fix some more global-related problems in other modules. Do you prefer several small commits or should I squash them?

diegonehab:
Small thematic fixes are better because we can discuss them independently and merge them when we are happy.

mpeterv:
Pushed them all.

diegonehab:
Let me see.

diegonehab:
Thanks!

mpeterv:
No problem!

PR #161, opened 2016-02-16
Rewrite try/protect in Lua, support table errors

mpeterv:


diegonehab:
Merged  #162 that implements it in C.

diegonehab:
Do

mpeterv:
Probably not, there are a few implementations available on Lua wiki and elsewhere already.

ignacio:
Would it be worth it to leave those links here for posterity?

diegonehab:
I think so. Particularly because of the new idea of checking the private metatable to make sure the error was thrown by try(). Do these other Lua implementations use the same idea?

moteus:
In my library i add special [unique key](https://github.com/moteus/lua-try/blob/master/src/try.c#L57) to error wrap table.

Issue #156, opened 2016-01-25
Incorrect/incomplete docs for socket.try and protect


mpeterv:
It should be mentioned that socket.try wraps errors in a table (right now it says `otherwise, it calls error passing ret2` which is wrong). Likewise, it should be mentioned that socket.protect considers all tables "exceptions". (The fact that socket.protect silently drops tracebacks and that tables as errors are not supported is another bug?)

diegonehab:
Can you post a pull request fixing the issue? Perhaps the best way to treat the problem would be to create a metatable, set it as upvalue to protect and try, then use the upvalue to set the metatable in wrap and check in unwrap.

mpeterv:
@diegonehab would you accept it if I rewrote the try/protect module in Lua? It would be much clearer, I think, and need for lua_pcallk hacks would be gone. Considering luasocket's job is to do expensive networking syscalls, performance cost is negligible.

siffiejoe:
@mpeterv: I already [tried once](https://github.com/diegonehab/luasocket/issues/112#issuecomment-62415654) ...

diegonehab:
What exactly is the proposal? We need support for Lua 5.1, 5.2, and 5.3. I'm a bit rusty, but I believe the only way to make it work across a yield barrier in Lua 5.1 is to implement it in Lua using the complicated coroutine hack. Starting in 5.2, the C version can be used across yield barrier and the code is quite simple. It can be made more robust with the metatable idea I suggested. It can also be implemented in Lua and it would also be quite simple. I don't know that the performance impact would be, but am to agree that it wouldn't be too bad. We'd still have the problem with Lua 5.1, right?

mpeterv:
The yield barrier problem is orthogonal to the original issue, and I don't have a strong opinion on whether luasocket should use coxpcall to work around it. My proposal is to do this, and do it in Lua for clarity:

> Perhaps the best way to treat the problem would be to create a metatable, set it as upvalue to protect and try, then use the upvalue to set the metatable in wrap and check in unwrap.

siffiejoe:
[...]

PR #162, opened 2016-02-21
Support table errors in LuaSockets LTN13 C implementation.

siffiejoe:
LuaSocket wraps error messages raised by newtry() in a table and unpacks them later so that (string) errors raised by 3rd-party code can be passed through as-is. This obviously didn't work when the 3rd-party code raised a table as an error message. This change sets a private metatable on all wrapped LuaSocket exceptions to distinguish them from 3rd-party table errors.

Since it is not yet decided whether to go with the current C implementation or rewrite the exception mechanism in Lua, this is an alternative pull request for the changes in #161 but in C instead of Lua.
See also #156 for the discussion.

### DIFF
--- a/doc/socket.html
+++ b/doc/socket.html
@@ -220,13 +220,6 @@ Returns an equivalent function that instead of throwing exceptions,
 returns <tt><b>nil</b></tt> followed by an error message. 
 </p>
 
-<p class=note>
-Note: Beware that if your function performs some illegal operation that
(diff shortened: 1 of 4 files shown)


diegonehab:
This is great siffiejoe! The only thing I would change is to keep the upvalue with the metatable for the errors as upvalue(1) in all functions.

diegonehab:
Your idea of using call instead of pcall for the finalizer makes sense as well...

PR #170, opened 2016-04-09
URL-decode user password before adding to authorization header.

bytefire:
If a URL like `http://username:p%40ssword@server.com/` - where @ symbol in password is URL-encoded - is passed to `http.request()` method, the password is added to Authorization header as it is, leading to failed authentication response from server. Decoding the password just before adding it to the header resolves this issue. That is what this pull request does.

This should also resolve same problem in LuaSec's `https.request()` function will depends upon LuaSocket.

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -222,7 +222,8 @@ local function adjustheaders(reqt)
     -- if we have authentication information, pass it along
     if reqt.user and reqt.password then
         lower["authorization"] =
-            "Basic " ..  (mime.b64(reqt.user .. ":" .. reqt.password))
+            "Basic " ..  (mime.b64(reqt.user .. ":" ..
+		url.unescape(reqt.password)))
     end
     -- if we have proxy authentication information, pass it along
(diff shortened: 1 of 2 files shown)


diegonehab:
Good catch! Please use url.unescape() instead, for clarity. I'll accept the pull request once you do that.

bytefire:
Sure. And update the comment on top of `url.escape()` as well?

diegonehab:
The comment seems reasonable for either version of the code. Up to you. Thanks!

bytefire:
Comment on _M.unescape() in url.lua says 

-- Encodes a string into its escaped hexadecimal representation

That's what made me think again initially :)

diegonehab:
Ah! I thought you meant your own comment. Please fix the code in url.lua as well. Sorry about that. :)

bytefire:
Right, I've ended up reverting and re-adding the changes instead of updating the previous pull-request. Hope that's okay. Let me know if the changes don't make sense and I'll update them. Cheers!

Issue #165, opened 2016-02-23
Add UDP to unix domain socket


dzuy65:
This is more like a request.  Have you considered addind UDP to unix domain socket?  I'm willing to contribute,

diegonehab:
Sure, we could have unix.tcp() and unix.udp(). How useful is this? I don't mind adding the support. I never get much feedback from the unix domain implementation.

dzuy65:
Since lua socket has tcp, udp, and unix, unix.tcp, it would just round up the features with unix.udp.
I can use it for my application actually.  I have 2 daemons that use unix socket, but they would lock up some times because on could wait for the other which is blocked by tcp socket.  It would be nice to have unix.udp to avoid blocking.

diegonehab:
So the idea is to split unix.c into unixtcp.c and unixudp.c. Then have unix.c init both modules, create a "unix" table, load both tcp() and udp() into the table, and return it. Right?

dzuy65:
I'm not familiar with the code so I can't really comment on it, but I know that unix.c assumes TCP by default.  It would be nice to have a UDP option.  Yes, I think there should be a unix domain socket, then a choice to use TCP or UDP, with TCP as default.

diegonehab:
Right now, unix.c is kind of an experimental beast. It builds into its own library, and I am not sure how people use it. I suppose they run something like 

    local socket = require"socket"
    socket.unix = require"socket.unix"
    u = socket.unix()

This creates a stream socket.

We  should change this to something like

    local socket = require"socket"
    socket.unix = require"socket.unix"
    u1 = socket.unix.tcp()
    u2 = socket.unix.udp()

Each call creating the appropriate type of socket.

The proper way to do this would be to keep in unix.c only the functions that are common to tcp and udp. Then create a unixtcp.c and move the tcp code to it. Create unixudp.c and write udp code into it. Then add a luaopen_unix in unix.c that creates and returns a table with the tcp and udp functions implemented in unixtcp.c and unixudp.c.

Does that make sense?

dzuy65:
Yes, I'm using 

u = socket.unix()

Reading the code, looks like the unix.c code has all the TCP methods, some may be common to UDP.
I'm thining of:

usocket = socket.unit(proto)

where proto can be either "udp" or "tcp", then you'd just add more UDP methods to unix.c without adding new files.

But you know your code best.  I'll go with whatever you design.

diegonehab:
[...]

PR #181, opened 2016-06-24
Add support for datagram unix domain sockets

enginix:
We use datagram unix domain socket as the IPC mechanism in our project, and I found datagram sockets not supported here. So I added the support for it, follow your suggestion in #165 . Test OK on macosx, with test/unix*.lua.

Thanks for your work of this project.


diegonehab:
Thank you for the code!

PR #183, opened 2016-07-05
Handle port as number in socket.url.build

mpeterv:
In particular, luasec passes port as number to socket.url.build.
Under Lua 5.3 this can lead to '.0' added to port when converting to
string (or an error if implicit conversions are disabled), use 'string.format("%d", port)' to avoid that.


mpeterv:
Sorry, of course this is an issue of luasec, it's documented that port should be a string.

PR #190, opened 2016-09-19
Fix for slow tcp connect in non blocking mode

tdev:
Hotfix for nonblocking, slow TCP connections not returning instantly. If the connect() takes too long, it will return WSAEALREADY, which when linked incorrectly, return WSAINVAL instead. This will destroy the connection.

Quote from: https://msdn.microsoft.com/en-us/library/windows/desktop/ms737625(v=vs.85).aspx

> WSAEALREADY: A nonblocking connect call is in progress on the specified socket.
> Note  In order to preserve backward compatibility, this error is reported as WSAEINVAL to Windows
> Sockets 1.1 applications that link to either Winsock.dll or Wsock32.dll."

The attached patch converts any WSAINVAL to WSAEALREADY. This has the problem, that if you really have invalid arguments (call connect on a listening socket) it will not fatally fail anymore.

### DIFF
--- a/src/wsocket.c
+++ b/src/wsocket.c
@@ -125,7 +125,19 @@ int socket_connect(p_socket ps, SA *addr, socklen_t len, p_timeout tm) {
     if (connect(*ps, addr, len) == 0) return IO_DONE;
     /* make sure the system is trying to connect */
     err = WSAGetLastError();
+
+    if (err == WSAEISCONN) return IO_DONE; // already connected?
+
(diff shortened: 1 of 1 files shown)


ewestbrook:
Can this be done without breaking the listening socket error case?

In other words, can we test whether it's a listening socket and only convert `WSAINVAL` to `WSAEALREADY` when it's not?

If we can't test for that, should we make note of it when `socket_listen` is called?

Issue #193, opened 2016-11-24
problem with unix socket


devbp:
I am trying to connect to unix scoket server. I have lau 5.2.3- when I use the code below, I get error

attempt to call field 'unix' (a table value)
stack traceback:

socket = require"socket"
socket.unix = require"socket.unix 
c = assert(socket.unix())
assert(c:connect("/tmp/foo"))
    while 1 do
        local l = io.read()
        assert(c:send(l .. "\n"))
    end_

Does lua 5.2.3 support unix socket in lua?

diegonehab:
Yes. We now support stream and dgram methods. The current names are "udp()" and "tcp()" and they live in the table you received in your call to require(), though we should probably change them to "dgram()" and "stream()".

diegonehab:
Maybe @enginix can change it? What do you think?

enginix:
Agree, "dgram()" and "stream()" are better here. I'll change it in a few days, as i am quite busy recently.

devbp:
this works for me now..
local socket = require"socket"
socket.unix = require"socket.unix"
u1 = socket.unix.tcp()
u2 = socket.unix.udp()

assert(u1:connect("./sock"))
    while 1 do
        local l = io.read()
        assert(u1:send(l .. "\n"))
    end

diegonehab:
Good. No hurry. Thanks!

tkim5574:
attempt to index a function value (field 'unix')

The above is the error I get. Is this changed since the patch?

diegonehab:
Works for me.

    Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
    > unix = require"socket.unix"
    > a = unix.dgram()
    > print(a)
    unixdgram{unconnected}: 0x1cf8188

tkim5574:
>Lua 5.3.4  Copyright (C) 1994-2017 Lua.org, PUC-Rio
>-socket._VERSION
LuaSocket 3.0-rc1
>-unix = require ('socket.unix')
>-unix
function: 0x7fe50850b450
>-a = unix.dgram()
stdin:1: attempt to index a function value (global 'unix')
stack traceback:
	stdin:1: in main chunk
	[C]: in ?

tkim5574:
nvm! There was a problem during the installation. Installing with 'make install-both-unix' worked for me

PR #199, opened 2016-12-25
Rename unix.tcp to unix.stream, unix.udp to unix.dgram

enginix:
Changes for what we discussed in #193


sathieu:
This is hard to follow. You're changing the API too often, and this breaks Prosody (See https://bugs.debian.org/852250).

diegonehab:
The unix socket implementation has always been kind of a curiosity. Similar to the serial implementation. They are not even documented. In that sense, the API is not official. So I didn't feel very strongly about maintaining compatibility. I guess the best solution is to have a deprecation policy and have unix.tcp = unix.stream and unix.udp = unix.dgram temporarily. What do you say?

sathieu:
What I need is socket.unix=socket.unix.tcp=socket.unix.stream, but is it possible?

cbane:
I just opened a pull request (https://github.com/diegonehab/luasocket/pull/207) that adds backwards compatibility aliases and wrappers.

diegonehab:
Sometimes the flexibility of Lua is a bit scary. Very nicely done.

PR #206, opened 2017-01-25
Create socket on first sendto if family agnostic udp() was used

ncopa:
Create socket and set family on first sendto() if udp() was created
without address family.

Signed-off-by: Natanael Copa <ncopa@alpinelinux.org>


ncopa:
fixed typo in commit message

diegonehab:
Thanks!!!

PR #207, opened 2017-01-26
Add backwards compatibility wrappers for socket.unix

cbane:
Add backwards compatibility aliases "tcp" and "udp" for the recently renamed "stream" and "dgram" functions, as well as a wrapper function and metatable setup so that socket.unix() calls socket.unix.stream().

If desired, I can update this so that the backwards compatibility versions print a deprecation warning on first use.

### DIFF
--- a/src/unix.c
+++ b/src/unix.c
@@ -17,14 +17,54 @@ static const luaL_Reg mod[] = {
     {NULL, NULL}
 };
 
+static void add_alias(lua_State *L, int index, const char *name, const char *target)
+{
+    lua_getfield(L, index, target);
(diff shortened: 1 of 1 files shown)


sathieu:
Thanks @cbane

sathieu:
It looks like it's not enough (at least for lua5.1), as previously socket.unix was global. The following was setting socket.unix to the function:
```
local socket= require "socket"
pcall(require, "socket.unix")
```

cbane:
I have some changes that make a plain `require("socket.unix")` call register `socket.unix`. Should I open another pull request for that?

PR #215, opened 2017-05-10
Update usocket.c

isaacgraf:
In our application we find that under certain conditions, SIGPIPE signals the occur when socket is writing kills the process, even though the signal is ignored in the socket_open function.
I found that adding the SIG_IGN handler for the signal before each write call to the socket fixed the issue for us.

### DIFF
--- a/src/usocket.c
+++ b/src/usocket.c
@@ -200,6 +200,8 @@ int socket_send(p_socket ps, const char *data, size_t count,
     *sent = 0;
     /* avoid making system calls on closed sockets */
     if (*ps == SOCKET_INVALID) return IO_CLOSED;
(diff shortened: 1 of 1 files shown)


diegonehab:
LuaSocket would be a bit rude to mask SIGPIPE for the entire application, as it is unclear what interactions this would cause. Can you use SO_NOSIGPIPE in your system instead?

diegonehab:
Actually, I just checked and socket_open already masks the signal. What gives? Is it not being called? Is some other library reseting the handler?

isaacgraf:
Thank you very much for your response. I looked into it further and indeed there was another library that was resetting the handler.  I will close the pull request.

PR #221, opened 2017-08-24
Fix include directory on Makefile

gut:
The lua.h file is found under /usr/include/lua5.1/ and not under
/usr/include/lua/5.1/ on my Ubuntu linux machine (X64 and PPC64le
architectures).


ewestbrook:
Both forms are actually needed in order to accommodate all known distro patterns.  See #265.

PR #223, opened 2017-08-30
fix buffer overflow in opt_meth_setoption

QuanTech0:
PoC trigger:
c=socket.connect('google.com',80)
c.setoption(c,'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -3,6 +3,7 @@
 * LuaSocket toolkit
 \*=========================================================================*/
 #include <string.h>
+#include <stdlib.h>
 
 #include "lauxlib.h"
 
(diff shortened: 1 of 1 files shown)


src/options.c
```diff
@@ -37,9 +38,10 @@ int opt_meth_setoption(lua_State *L, p_opt opt, p_socket ps)
     while (opt->name && strcmp(name, opt->name))
         opt++;
     if (!opt->func) {
-        char msg[45];
+        char* msg = malloc(30+strlen(name));
         sprintf(msg, "unsupported option `%.35s'", name);
         luaL_argerror(L, 2, msg);
+        free(msg);
```
moteus: This is memory leak. Use push string, free and then error.

siffiejoe: `lua_pushfstring()` would be a leak-free (and more convenient) alternative. Or just make the buffer large enough -- at least 56 characters if I counted correctly.

ewestbrook:
Fixed in #239.

PR #238, opened 2017-11-19
fixed URL parsing in url.lua: parse fragment after parsing username a…

hleuwer:
…nd password.

### DIFF
--- a/src/url.lua
+++ b/src/url.lua
@@ -139,14 +139,14 @@ function _M.parse(url, default)
         parsed.authority = n
         return ""
     end)
-    -- get query string
-    url = string.gsub(url, "%?(.*)", function(q)
-        parsed.query = q
-        return ""
-    end)
     -- get fragment
     url = string.gsub(url, "#(.*)$", function(f)
         parsed.fragment = f
+        return ""
+    end)
+    -- get query string
+    url = string.gsub(url, "%?(.*)", function(q)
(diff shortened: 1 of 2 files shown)


diegonehab:
Thanks for the contribution. Can you post an example of URL that was couldn't be parsed before but that now can be parsed correctly?

diegonehab:
Can you modify your pull request to include a few test cases that showcase the modification in testurl.lua? Then I can just merge the request.

hleuwer:
Thanks for your quick reaction!
I didn't perform the url tests before placing the pull request. Now I did. 
I committed a new version to my fork of luasocket:
- postpone fragment parsing just after authority parsing before query parsing
- added a couple of test cases for passwords with special characters to test/urltest.lua
Sorry for the confusion.
However, I found another problem in the test cases for check_protect(...):
- reference patterns should contain upper case hex letters - the check_protect(...) test cases fail otherwise.

diegonehab:
You mean that the upper case hex letters were not being tested but worked fine anyway, right? Otherwise, so we need to fix something else?

hleuwer:
No, they were tested but the tests failed to my surprise - even on a freshly cloned version. The function to build the url from special characters uses the correct formatting pattern %0X, but the reference pattern for comparison were lower case, which made the tests fail. So I changed the reference pattern and all tests pass.

Issue #242, opened 2018-03-08
use of undeclared identifier 'EAI_OVERFLOW' on Mac OS X 10.5


mojca:
Compiling TeX Live for Mac OS X 10.5 fails:
```
In file included from ../../../source/texk/web2c/luatexdir/luasocket/src/socket.c:4:
../../../source/texk/web2c/luatexdir/luasocket/src/usocket.c:441:14: error: use of undeclared identifier 'EAI_OVERFLOW'
        case EAI_OVERFLOW: return PIE_OVERFLOW;
             ^
```

See also https://github.com/italiangrid/voms/issues/2

(Compiled with clang 5.0.)

mojca:
Below is the complete content of the `netdb.h` library on 10.5. I tried defining `_POSIX_C_SOURCE`, but then I get an equally useless result/error:
```
../../../texk/web2c/luatexdir/luasocket/src/inet.c:84:15: error: use of undeclared identifier 'NI_MAXHOST'
    char hbuf[NI_MAXHOST];
              ^
../../../texk/web2c/luatexdir/luasocket/src/inet.c:85:15: error: use of undeclared identifier 'NI_MAXSERV'
    char sbuf[NI_MAXSERV];
              ^
../../../texk/web2c/luatexdir/luasocket/src/inet.c:141:58: error: no member named 'h_addr' in 'struct hostent'
    lua_pushstring(L, inet_ntoa(*((struct in_addr *) hp->h_addr)));
                                                     ~~  ^
../../../texk/web2c/luatexdir/luasocket/src/inet.c:179:19: error: use of undeclared identifier 'NI_MAXHOST'
        char hbuf[NI_MAXHOST];
                  ^
```

Note that the two apparently cannot be defined at the same time:
```c
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
// ...
#else  /* (_POSIX_C_SOURCE && !_DARWIN_C_SOURCE) */
#define EAI_OVERFLOW    14	/* An argument buffer overflowed */
#endif /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */

#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
#define	NI_MAXHOST	1025
#define	NI_MAXSERV	32
#endif /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */
```

Library:
```c
#ifndef _NETDB_H_
[...]

PR #243, opened 2018-03-21
EAI_OVERFLOW, AI_NUMERICSERV: put behind #ifdef

mojca:
It may not be defined on some systems.

See some other comments in #242

There's a remaining issue that `AI_NUMERICSERV` from `src/udp.c` is also not defined. Not sure how to fix that one other than also commenting it out.

### DIFF
--- a/src/udp.c
+++ b/src/udp.c
@@ -188,7 +188,10 @@ static int meth_sendto(lua_State *L) {
     memset(&aihint, 0, sizeof(aihint));
     aihint.ai_family = udp->family;
     aihint.ai_socktype = SOCK_DGRAM;
-    aihint.ai_flags = AI_NUMERICHOST | AI_NUMERICSERV;
+    aihint.ai_flags = AI_NUMERICHOST;
+#ifdef AI_NUMERICSERV
+    aihint.ai_flags |= AI_NUMERICSERV;
+#endif
     err = getaddrinfo(ip, port, &aihint, &ai);
 	if (err) {
         lua_pushnil(L);
--- a/src/usocket.c
+++ b/src/usocket.c
@@ -440,7 +440,9 @@ LUASOCKET_PRIVATE const char *socket_gaistrerror(int err) {
         case EAI_FAMILY: return PIE_FAMILY;
         case EAI_MEMORY: return PIE_MEMORY;
         case EAI_NONAME: return PIE_NONAME;
+#ifdef EAI_OVERFLOW
         case EAI_OVERFLOW: return PIE_OVERFLOW;
+#endif
 #ifdef EAI_PROTOCOL
         case EAI_PROTOCOL: return PIE_PROTOCOL;
 #endif


diegonehab:
Any idea of what system we are talking about here? Also for the case of AI_NUMERICSERV?

mojca:
> Any idea of what system we are talking about here?

More details are in #242.
This broke for me on Mac OS X 10.5, so we patched it locally.

diegonehab:
I guess we can add the relevant ifdefs, I don't have a way to test this. Is 10.5 still in use?

mojca:
I'm not claiming it's the most popular platform around, and I doubt it's used as the main machine anywhere (some do still use 10.6 as their main OS as they didn't like the direction where Apple was heading afterwards), but it's the last OS that works on PPC, so if you don't want to turn a 20 kg Mac Pro into a brick, that's what you are left with. We have some geeks who submitted & merged 7 pull requests for Tiger (10.4) support just in the last three days (which beats me as well, but hey ...)

I can try to test or ask @kencu for help, but it would be helpful to know what specifically to test. My initial main intent was to build the LuaTeX binaries in the first place, and I have no way of knowing whether LuaTeX even uses that part of code.

I tried to run the tests from the package manager now, but I have one tiny issue. We usually run the tests before installing the packages (that's a typical workflow, so that you don't install broken software). Basically we run

```Makefile
make DEBUG=DEBUG macosx LUAV=5.3
make DEBUG=DEBUG LUAV=5.3 DESTDIR=/some/strange/temp/path PREFIX=/opt/local PLAT=macosx install-unix
```

but then the tests refuse to run since there's no code in place to tell lua to look for the libraries in `/some/strange/temp/path`. This would be nice to fix (either on your side, or we can fix it on our side if you tell us how).

For the tests on 10.5 we can exceptionally try to first install the library first and then run the tests, the only question is whether the unit tests were supposed to catch potential failures related to missing definitions at hand.

diegonehab:
I sympathize.

Testing before installation is possible, but maybe it is easiest to install to a temp directory and point LUA_PATH and LUA_CPATH there?

When I said this needed testing, I meant I can't test to see if it even builds after we make the changes.

kencu:
Hello. I have lots of these systems, and I'd be happy to work with Mojca to get it building and run the tests.

diegonehab:
The AI_NUMERICHOST and AI_NUMERICSERV avoid a potentially slow name resolution for the hostname (which will always be an ip address) and for service name (which will always be an actual port number).  By omitting any of these, the worst thing that can happen is the code becoming slower. Should still work, though. I can't test the impact on speed.

Thanks @kencu. Can you check that else we need in order for LuaSocket to compile properly on whatever version of Mac OS your users need? Then we add a commit to this pull request and solve everything in one go.

kencu:
Yeah, well good/bad news. The current MacPorts `lua-luasocket` port installs without any modifications on 10.4 PPC and 10.5 PPC, and all the tests in the test suite pass.

@mojca -- if there is some newer version of the `lua-luasockets` git commit you were hoping to test, let me know.

mojca:
> Yeah, well good/bad news. The current MacPorts `lua-luasocket` port installs without any modifications on 10.4 PPC and 10.5 PPC

You mean: after you apply these patches, right?
https://github.com/macports/macports-ports/tree/master/devel/lua-luasocket/files

kencu:
Ah. I didn't notice that the port was already patched in Macports, but certainly it is. Yes, with the current Macports' patches, it builds and tests properly on 10.4 & 10.5.

mojca:
I have shamelessly stolen the other patch from MacPorts (I probably never payed attention to our existing patches).

@kencu: you say that you have the tests working. Can you please modify the port, so that `port test` will work automatically? I was probably missing some variable to point to the location of files before being installed.

Question to the authors: is there any plan to get a new official release out at some point? TeX Live release cycle is now quickly approaching. If a new release of luasocket gets out soon, we could still include it.

diegonehab:
@mojca: We are trying to close as many of the outstanding issues as possible.

@kencu: Thanks. So I will merge here.

diegonehab:
@kencu: I have another unrelated question. Lately, the 

    export MACOSX_DEPLOYMENT_TARGET="10.3"

began causing issuing a warning on recent versions of Mac OS X. Can I simply remove it from the makefile?

kencu:
Yes. By default, all macOS build systems I know of set the deployment target to the system they are building on, which is appropriate 99% of the time. 

Someone who wants to do something fancy like deploy on an older system than the build system will need to know what they are doing, so would not get caught by this, IMHO.

mojca:
> ```
> export MACOSX_DEPLOYMENT_TARGET="10.3"
> ```
> began issuing a warning on recent versions of Mac OS X. Can I simply remove it from the makefile?

Absolutely, for at least two reasons:
* You don't need this set at all, unless you want to build binaries yourself, targeting Mac OS X 10.3. You don't want that.
* 10.3 is totally obsolete. Trying to install a decent compiler is a nightmare and the set of functions available out of the box doesn't suffice to build anything reasonable.
(I didn't notice this one.)

mojca:
And many many thanks for the fix.

diegonehab:
> Question to the authors: is there any plan to get a new official release out at some point? TeX Live release cycle is now quickly approaching. If a new release of luasocket gets out soon, we could still include it.

We are working for it. How soon are you looking at?

mojca:
Here's the timeline: https://www.tug.org/texlive/
The 22nd February was officially the deadline to commit "simply anything", the sources were supposed to be ready by now, but the binaries were supposed to be committed by the 1st of March. The 22nd March is the final freeze. The sooner the freeze date, the less likely the changes are to be accepted.

But don't worry about this.

diegonehab:
Give us a ping when the deadline is closer. We can, at least, tag a version, right? It won't be worse than the current situation. :)

PR #244, opened 2018-03-25
add options:

leyyer:
get/set : recv-buffer-size
get/set : send-buffer-size

Signed-off-by: surenyi <surenyi82@163.com>


leyyer:
I use udp socket, discovery always dropping packet, so add these options.

PR #249, opened 2018-06-03
Update auxiliar.c

zhaozg:
auxiliar_getclassudata should return NULL but generate errors


diegonehab:
Why should it return NULL instead or raising an error?

zhaozg:
To @diegonehab 
https://github.com/zhaozg/luasocket/blob/c57848ca285eb7a224bf16bf81e551207f8a24cb/src/auxiliar.c#L86-L90 will raise error, but https://github.com/zhaozg/luasocket/blob/c57848ca285eb7a224bf16bf81e551207f8a24cb/src/auxiliar.c#L141-L145 should return NULL

zhaozg:
Please use #250

PR #250, opened 2018-06-05
Update auxiliar.c to use luaL_testudata (#249), now with Lua 5.1 compatibility

ewestbrook:
Provides PR #249 with a copy of `luaL_testudata` for the Lua 5.1 build case.

### DIFF
--- a/src/auxiliar.c
+++ b/src/auxiliar.c
@@ -143,7 +143,7 @@ void *auxiliar_getgroupudata(lua_State *L, const char *groupname, int objidx) {
 * otherwise
 \*-------------------------------------------------------------------------*/
 void *auxiliar_getclassudata(lua_State *L, const char *classname, int objidx) {
-    return luaL_checkudata(L, objidx, classname);
+    return luaL_testudata(L, objidx, classname);
 }
 
 /*-------------------------------------------------------------------------*\
--- a/src/compat.c
+++ b/src/compat.c
@@ -16,4 +16,22 @@ void luaL_setfuncs (lua_State *L, const luaL_Reg *l, int nup) {
   }
   lua_pop(L, nup);  /* remove upvalues */
 }
+
+/*
+** Duplicated from Lua 5.2
+*/
+void *luaL_testudata (lua_State *L, int ud, const char *tname) {
+  void *p = lua_touserdata(L, ud);
+  if (p != NULL) {  /* value is a userdata? */
+    if (lua_getmetatable(L, ud)) {  /* does it have a metatable? */
+      luaL_getmetatable(L, tname);  /* get correct metatable */
+      if (!lua_rawequal(L, -1, -2))  /* not the same? */
+        p = NULL;  /* value is a userdata with wrong metatable */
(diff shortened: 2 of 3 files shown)


diegonehab:
Why do you propose this change? It's supposed to raise an error when the object is not a userdata, no?

ewestbrook:
Frankly, I don't know -- I just nudged it to build for Lua 5.1, because it didn't.

@zhaozg, why indeed?

diegonehab:
It builds here fine with Lua5.1 without this modification.

ewestbrook:
Maybe I'm confused.  I'm building with `make LUAINC_linux=/usr/include LUAPREFIX_linux=/usr PLAT=linux LUAV=5.1` here.

Since `luaL_testudata` doesn't exist in Lua 5.1, using commit e1e41be948196a98af67345d51fe7a238c850c54 (@zhaozg's original, before I added the function to `compat.c`), I get an implicit declaration for the missing function ("just" a compiler warning, but certainly a harbinger):
```
auxiliar.c:146:12: warning: implicit declaration of function ‘luaL_testudata’; did you mean ‘luaL_checkudata’? [-Wimplicit-function-declaration]
     return luaL_testudata(L, objidx, classname);
            ^~~~~~~~~~~~~~
            luaL_checkudata
```
And then the "actual" error comes as an undefined symbol when the runtime tries to load it:
```
/usr/bin/lua: error loading module 'socket.core' from file '/usr/lib/lua/5.1/socket/core.so':
        /usr/lib/lua/5.1/socket/core.so: undefined symbol: luaL_testudata
stack traceback:
        [C]: ?
        [C]: in function 'require'
        /usr/share/lua/5.1/socket.lua:12: in main chunk
```
But if it's not the better call to use in the first place, this is all rather moot, I guess.

diegonehab:
Oh, you mean it doesn't build if you simply change luaL_checkudata to luaL_testudata, right?
@zhaozg is right that auxiliar_getclassudata will raise an error if it is called on a object that is not a userdata or not a member of the class. In Lua 5.0, it would return NULL, but this behavior changed with Lua 5.1 and I did not realize it. However, that function is only called in auxiliar.c itself, by auxiliar_checkclass, and that function will raise an error anyway when auxiliar_getclassudata returns NULL. So this "bug" can never manifest itself in LuaSocket. You can't simply change luaL_checkudata to luaL_testudata because Lua 5.1 does not have luaL_testudata. But why do I have two pull requests with the same code change?

ewestbrook:
Right!  Sorry not to have been clearer.  I'd also like to apologize for the duplicate PR, having learned that my commit's appearance in #249 made it unnecessary to create #250 in the first place.  There should only be one PR for this.  Please feel free to close if you wish.

diegonehab:
Alright. I'll merge his PR and close this one. Can you test if it works for you?

zhaozg:
Thank @diegonehab  and @ewestbrook

ewestbrook:
Tests good here.  Thanks again!

Issue #254, opened 2018-08-17
url.absolute() can alter hostname when applying relative path


dev-random:
```lua
local url = require 'socket.url'
print(url.absolute('http://example.com/', '../.badhost.com/')
```
Expected: either ```http://example.com/../.badhost.com/``` or ```http://example.com/.badhost.com/```

Got: ```http://example.com.badhost.com/```

ewestbrook:
Fixed in #256.

dev-random:
Stil broken:
```lua
url.absolute("http://example.com/", "...badhost.com")
```
Got: ```http://example.com.badhost.com/```

ewestbrook:
It's still unmerged; could that be why you're not seeing it?  Checking out the branch in progress produces the correct result for me here.

dev-random:
"ewestbrook/luasocket/luasocket254"? That's what I'm testing.
Maybe I should've made it more clear that the test cases in the two posts differ? You did fix the issue in the original post, but introduced an almost identical issue described in the reply.

ewestbrook:
Ah!  Thank you, I overlooked the subtle difference in your second example.  Investigating.

ewestbrook:
Better now, and thanks for testing.  I've also added your examples verbatim to the self-tests to help avoid future regression.

dev-random:
```lua
print(url.absolute("http://example.com/a/b/c/d/", "../../../../q"))
```
Expected: ```http://example.com/q```
Got: ```http://example.com/a/b/q```

ewestbrook:
Got it.  Thanks again, keep 'em coming!

dev-random:
```lua
print(url.absolute("http://example.com", ".badhost.com"))
```
Expected: ```http://example.com/.badhost.com```
Got: ```http://example.com.badhost.com```

ewestbrook:
Nice catch. Fixed as well.

dev-random:
A corner case:
```lua
print(url.absolute("http://example.com/a/b/c/d/", "..//../../../q"))
```
Expected: ```http://example.com/a/q```
Got: ```http://example.com/a/b/c/d/..//../q```

Empty path components are valid components that ".." should be able to remove.

ewestbrook:
Tricky, and indeed, apparently legal.  Fixed.

dev-random:
```lua
print(url.absolute("http://example.com/a/b/c/d/", "..//a/../../../../q"))
```
[...]

PR #256, opened 2018-08-19
url.lua:absolute_path(): fix issue #254, simplify, add more test cases

ewestbrook:


ewestbrook:
This fix is intended to cover the present case without breaking [others](https://github.com/diegonehab/luasocket/blob/88b13a825b6c514d243272d3fc598a4ba56ebe3e/test/urltest.lua#L620).

The current parsing code is documented as having been written against [RFC 2396](https://tools.ietf.org/html/rfc2396), which of course has since been superceded by [RFC 3986](https://tools.ietf.org/html/rfc3986).  The latter has more guidance for cases like this and others, so I'm experimenting with larger changes to more closely follow it.  But for now this should do.

diegonehab:
It is great to see a pull request that not only changes code, but also changes the tests.
Do you have an idea of how much additional code would be needed to fully support RFC 3986? It would be best not to stop halfway.
For example, page 31 of the new RFC shows pseudo-code.
From there, I looks as though all we need is a remove_dot_components function.
Shall we follow the suggestion?
Then we can also use it in function _M.absolute.
In particular, the remove_dot_components must always be called, even when there is no merging of paths.

PR #264, opened 2019-02-19
src/unixdgram.c: fix connect() and bind() on FreeBSD

ewestbrook:
From #101. Allows connect() and bind() without dummy char.  Needs testing.

### DIFF
--- a/src/unixdgram.c
+++ b/src/unixdgram.c
@@ -17,6 +17,12 @@
 
 #define UNIXDGRAM_DATAGRAMSIZE 8192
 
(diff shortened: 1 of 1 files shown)


diegonehab:
I think that, if we are going to make this kind of change, we might as well use the SUN_LEN macro?
Then, if there is a sun_len field we set it t o SUN_LEN(&local)+1, and in the call to connect or bind we pass SUN_LEN(&local). What do you think?

ewestbrook:
Good call.  Something like that?

diegonehab:
Yes!

PR #268, opened 2019-02-23
Scheme-independent connection and redirection

ewestbrook:
Automatically handles upgrade redirects from HTTP to HTTPS using LuaSec, and allows both HTTP and HTTPS requests to be made using the single `http.request()` function. Requires only the `tcp()` function to be exposed by `ssl.https`. Also provides a path for support of other schemes in the future. See also [brunoos/luasec#132](https://github.com/brunoos/luasec/pull/132).

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -26,15 +26,24 @@ _M.TIMEOUT = 60
 -- user agent field sent in request
 _M.USERAGENT = socket._VERSION
 
--- supported schemes
+-- supported schemes and their particulars
 local SCHEMES = {
-    http = { port = 80 }
-    , https = { port = 443 }}
+    http = {
+        port = 80
+        , create = function(t)
+            return socket.tcp end }
+    , https = {
+        port = 443
+        , create = function(t)
+          local https = assert(
+            require("ssl.https"), 'LuaSocket: LuaSec not found')
+          local tcp = assert(
+            https.tcp, 'LuaSocket: Function tcp() not available from LuaSec')
+          return tcp(t) end }}
 
 -- default scheme and port for document retrieval
 local SCHEME = 'http'
 local PORT = SCHEMES[SCHEME].port
-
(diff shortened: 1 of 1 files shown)


diegonehab:
Oops. I just merged your older pull request for omitting the port number in the Host: header and now there is a conflict here. Can you resubmit?

I never played with tls on Lua. How do we know if the host has a valid certificate? Is this all automatic?

ewestbrook:
Thanks! It was expected, as I wasn't sure which would go first. Resolving now.

ewestbrook:
Resolved, tested, and ready.

> I never played with tls on Lua. How do we know if the host has a valid certificate? Is this all automatic?

It is.  That's all done by LuaSec.

Before this patch, you had to make a different call separately yourself if it was HTTPS (`require'ssl.https'.request()`), or else construct and pass a `create()` function with LuaSec and pass it to LuaSocket -- again, you had to do that separately yourself for the case of HTTPS.

This patch just makes it transparent and easy.  Same call, same arguments, regardless of scheme.

```
$ lua <<<"print(require'socket.http'.request'http://westbrook.com/')"
        200     table: 0x55f49d96dcc0   HTTP/1.1 200 OK

$ lua <<<"print(require'socket.http'.request'https://westbrook.com/')"
        200     table: 0x563792096e00   HTTP/1.1 200 OK
```

My favorite part is that if you make an HTTP request and get redirected to an HTTPS URL, it just works.  (The example HTTP request above got a 307 back and did exactly that, in fact.)

__Important__:  Please note that it does require the one line change in LuaSec PR 132 to expose the tcp() method there.  Hopefully that'll get merged on that side soon.

diegonehab:
Still says it has a conflict here. :(

ewestbrook:
That was weird. My local repo said it had already been merged, and this page said there was a conflict, and then it seemed to have sorted itself out.  Thanks.

Issue #139, opened 2015-05-12
Segmentation fault on CentOS


subnetmarco:
On CentOS 7 I experience a `segmentation fault` error when executing:

``` lua
local tcp = require("socket").tcp()
```

You can try with docker in interactive mode: `$ docker run -ti centos:7 /bin/bash`

diegonehab:
This is is most likely a build/installation problem, no? Anybody care to explain what's so different about CentOS that would cause LuaSocket to crash? I don't have access to a CentOS machine.

thefosk:
The problem was occurring when I was requiring luasocket inside of a function like:

```lua
local function test()
  local tcp = require("socket").tcp()
end
```

When moving the require outside of the function, everything works well:

```lua
local socket = require("socket")

local function test()
  local tcp = socket.tcp()
end
```

moteus:
May be you get segmentation fault when close lua_State?
It is possible that Lua unload library before close all sockets.

diegonehab:
What version of Lua are you using? During a "lua_close", Lua should GC all objects in the reverse order they are "created". So this wouldn't be a problem. Second, Lua should keep a reference to the library in the _LOADED table, so it won't be collected even though you didn't keep a reference to it. This is very strange.

thefosk:
I was using Lua `5.1.4` on OS X.

ewestbrook:
Is this problem still occurring?  Is there anything we can do to reproduce the problem or troubleshoot it further?

We're reviewing open items in preparation for a release.  If action is needed here, please add a comment.  Otherwise, this issue will be closed on or after 24-Feb-2019.

Thanks!

galtet:
My OS is Linux amzn2.x86_64, which is basically centos7, i get segmentation fault from the following code:
`local rex = require("rex_pcre2")
  local socket = require('socket').tcp()
`
The segfault backstrace:
`Backtrace:
/usr/lib64/liblua-5.1.so(lua_getallocf+0x9)[0x7fe207e59739]
/usr/local/lib/lua/5.1/rex_pcre2.so(Lmalloc+0x12)[0x7fe206cac702]
/usr/local/lib/lua/5.1/rex_pcre2.so(buffer_init+0x1e)[0x7fe206cac94e]
/usr/local/lib/lua/5.1/socket/core.so(+0x8ad5)[0x7fe20660aad5]
/usr/lib64/liblua-5.1.so(+0xc208)[0x7fe207e5d208]
[...]

PR #270, opened 2019-02-25
Tag functions explicitly for shared library visibility

ewestbrook:
Functions tagged, makefile updated, LuaRocks rockspecs updated.

### DIFF
--- a/luasocket-scm-0.rockspec
+++ b/luasocket-scm-0.rockspec
@@ -1,105 +1,132 @@
 package = "LuaSocket"
 version = "scm-0"
 source = {
-   url = "https://github.com/diegonehab/luasocket/archive/master.zip",
-   dir = "luasocket-master",
+  url = "git://github.com/diegonehab/luasocket.git"
+  , branch="master"
 }
 description = {
-   summary = "Network support for the Lua language",
-   detailed = [[
+  summary = "Network support for the Lua language",
+  detailed = [[
       LuaSocket is a Lua extension library that is composed by two parts: a C core
       that provides support for the TCP and UDP transport layers, and a set of Lua
       modules that add support for functionality commonly needed by applications
       that deal with the Internet.
    ]],
-   homepage = "http://luaforge.net/projects/luasocket/",
-   license = "MIT"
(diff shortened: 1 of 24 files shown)


ewestbrook:
Addresses #139

diegonehab:
I just merged the pull request, but didn't see the pragmas. It's not that we need them. We didn't even need to mark all internal symbols. I still think this is not a good solution. It is best to have the default  be hidden, and only mark explicitly the three symbols we want to export. The best way to do this is to use the compiler option. The second best is to use the pragmas, because one day somebody will add an internal function to LuaSocket and forget to add the LUASOCKET_PRIVATE decoration. If this happens and the default still is hidden, no problem. The user simply forgot to document the fact that he expects a given function to be private. But if the default is visible, then all hell breaks loose.

I think LuaRocks is wrong (twice). Once for not using this option by default for every library. And again for not using it for LuaSocket since we do that explicitly in the makefile. This problem *will* bite somebody else in the future. It may even bite us. :(

ewestbrook:
I agree completely.  I'm actually still working on the pragma approach.  (For some reason I'm getting a relocation error in the link phase.)  And I'm also not done trying to coerce LuaRocks into using `-fvisibility` without going all the way to `build.type='makefile'`.  But if we eventually have to go there, so be it; that'll still be better than this, for the reasons you just articulated.

diegonehab:
I never used the pragma myself. Maybe it changes something besides the visibility? when you compare the object file generated with -fvisibility=hidden and with no -fvisibility but with the pragma, do the symbols end up in a different section?

ewestbrook:
I'll look into that.  On a related topic, how do you feel about using `strip` during the `install` phase of the makefile?  It seems that distros commonly do that when they package builds.  Seems appropriate for non-debug builds.

diegonehab:
The binary is indeed smaller. 6x smaller sounds a lot, but the library is 360k before stripping... :) I don't think it makes much difference. 

In Mac OS X, I saw no difference. But removing the MACOSX_DEPLOYMENT_TARGET did reduce size.

PR #271, opened 2019-02-28
Use visibility pragma around declarations instead of attributes on definitions

ewestbrook:
Cleaner and a bit less error-prone implementation.  Still risks inadvertent exporting of new functions if declarations are not added to `.h` files along with others, but that should be a bit less likely than omitting the attributes on function definitions.


diegonehab:
Do these pragma get silently ignored by other compilers, or do they generate a lot of warnings? In other words, should we envelop them in some ifdefs?

ewestbrook:
I don't have any non-linux equipment to test on at this time.  I believe they are ignored -- but probably not silently.

I guess I should wrap in `#ifndef _WIN32` -- but are there others?

ewestbrook:
For what it's worth, I did confirm that this build achieves the goal under LuaRocks.

Testing steps for future reference:
- Edit `luasocket-scm-0.rockspec` to specify a unique branch with these changes;
- Push that unique branch to github;
- Pack a "source rock" with `luarocks --tree=tree pack luasocket-scm-0.rockspec`, which created `luasocket-scm-0.src.rock` from that branch;
- Build the code with `luarocks --tree=tree build luasocket-scm-0.src.rock`
- Test with `for i in $(find tree -name '*.so') ; do echo "===== $i =====" ; nm -Ca $i | grep ' T ' ; done`

The result was:
```
===== tree/lib/lua/5.1/socket/core.so =====
000000000000a1ac T _fini
00000000000033c0 T _init
0000000000003b40 T luaopen_socket_core
===== tree/lib/lua/5.1/socket/serial.so =====
0000000000005218 T _fini
00000000000019a0 T _init
00000000000051d0 T luaopen_socket_serial
===== tree/lib/lua/5.1/socket/unix.so =====
000000000000704c T _fini
0000000000002208 T _init
00000000000057f0 T luaopen_socket_unix
===== tree/lib/lua/5.1/mime/core.so =====
00000000000026b0 T _fini
0000000000000c90 T _init
00000000000023f0 T luaopen_mime_core
```
which is now the same result the makefile build produces.

src/auxiliar.h
```diff
@@ -31,7 +31,9 @@
 
 #include "luasocket.h"
 
+#ifndef _WIN32
```
siffiejoe: Shouldn't the `#ifndef` be GCC-specific?

Here is what I use in similar situations:
```C
#if (defined( __GNUC__ ) && __GNUC__ >= 4) || __has_attribute( __visibility__ )
```

You would need a fallback for all compilers that don't handle `__has_attribute`:
```C
#ifndef __has_attribute
#  define __has_attribute( x ) 0
#endif
```

siffiejoe: Although you probably have to handle `_WIN32 ` (-> MinGW) separately anyway ...

ewestbrook:
Rebased for latest master.

- Removes LUASOCKET_PRIVATE from function declarations;
- Adds pragma GCC visibility in .h files conditionally (when not _WIN32);
- Updates LuaRocks build to add missing compat.c in unix.so and serial.so builds;
- Removes some redundant include directives.

PR #273, opened 2019-03-02
add cmake

sonoro1234:
This is a cmake tested in mingw and nmake.
Only changes needed would be analogous to https://github.com/sonoro1234/luasocket/blob/cmakefor3_0/CMakeLists.txt#L8
to avoid compile not linux or not osx files (I am not sure which there are)

To test it just create a sibbling folder to luasocket repository folder and run
```
cmake  -G"Your desired generator"  -DCMAKE_BUILD_TYPE="Release"  -DLUA_INCLUDE_DIR="lua.h location" -DLUA_LIBRARY="lua library location" -DCMAKE_INSTALL_PREFIX="location of installation, generally where lua.exe is " ../luasocket
```

The list of available generators is in https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html

After it has finished generating just run

`make install` (for mingw32-w64 it would be `mingw32-make install`)

and thats all!!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,14 +10,10 @@ if(WIN32)
 	add_definitions(-D_WIN32_WINNT=0x0600) ##should be this or the next two for working on XP
 	#add_definitions(-D_WIN32_WINNT=0x0501)
 	#add_definitions(-DLUASOCKET_INET_PTON)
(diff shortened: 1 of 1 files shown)


sonoro1234:
The installation part was not enought for all.
Solution for my bundled repository in https://github.com/sonoro1234/Lua2SC/tree/master/luasocket_cmake
may be usefull for others?

PR #231, opened 2017-10-10
vs2012: added missing compat.c to project files

tdiff:

PR #275, opened 2019-03-11
Add src\compat.c to mime.vcxproj and socket.vcxproj

ewestbrook:
Rebases #231 against current master. Similar change was made to rockspecs in #271.

### DIFF
--- a/mime.vcxproj
+++ b/mime.vcxproj
@@ -20,6 +20,7 @@
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="src\mime.c" />
+    <ClCompile Include="src\compat.c" />
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{128E8BD0-174A-48F0-8771-92B1E8D18713}</ProjectGuid>
--- a/socket.vcxproj
+++ b/socket.vcxproj
@@ -21,6 +21,7 @@
   <ItemGroup>
     <ClCompile Include="src\auxiliar.c" />
     <ClCompile Include="src\buffer.c" />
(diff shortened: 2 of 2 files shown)


diegonehab:
Hmmm. I wonder why I didn't need compat when I tested on Windows. Seems like the kind of thing that would lead to a link error. Have you checked this?

ewestbrook:
I agree, and no, I haven't been able to check it on Windows myself.  Frankly, I'm not sure how it could link at all without it (on Lua 5.1 Windows builds, anyway).  I do know from my own testing that the LuaRocks build was broken until recently because, just like here, `compat` was missing from `mime` and `socket`.  However, that omission didn't come to light until I discovered it during the export pragma work. So perhaps something was obscuring the omission before then, or providing something else in its place until recently.  Wish I could be more helpful under Windows.

alerque:
To any Windows users who make come by...

What purpose exactly do these vcxproj files serve? Is there anyone around that is currently using them that could review this PR sensibly?

Tieske:
the `vcxproj` files are Visual Studio (that is full Visual Studio on Windows, not VS Code) project files, which contain the metadata on how to build a project etc, and all project/solution based settings.

looks fine to me.

Tieske (approved):

diegonehab:
Do we need the compat module, though?

Tieske:
it defines `luaL_setfuncs` and `luaL_testudata`, which were introduced in Lua 5.2. Both of those are used in multiple places in the codebase (mostly `luaL_setfuncs`)

diegonehab:
Makes sense. I forgot 5.1 is still widely used.

PR #280, opened 2019-04-21
SCM rockspec housekeeping

ewestbrook:


ewestbrook:
Done.  See also #280.  New rockspec is version scm-2 and available via [LuaRocks](https://luarocks.org/modules/luasocket/luasocket).

Please let me know if you are successful.  Appreciate your help!

Issue #297, opened 2020-03-20
socket tests indexed from global scope


jbbjarnason:
I am unable to run for example this test: https://github.com/diegonehab/luasocket/blob/master/test/tcp-getoptions#L3

```bash
jonb@ubuntu:~/luasocket/test$ ./tcp-getoptions 
lua: ./tcp-getoptions:21: attempt to index global 'socket' (a nil value)
stack traceback:
	./tcp-getoptions:21: in main chunk
	[C]: ?
```

is luasocket supposed to be accessible from the global scope?

ewestbrook:
Bug confirmed.  Opened #299 with partial fix.  Work in progress.

jbbjarnason:
I checked the other tests and found 3 other cases:
https://github.com/diegonehab/luasocket/blob/master/test/udp-zero-length-send
https://github.com/diegonehab/luasocket/blob/master/test/udp-zero-length-send-recv
https://github.com/diegonehab/luasocket/blob/master/test/find-connect-limit

ewestbrook:
Testing after applying the above patch:

Linux:

```
$ ./test/tcp-getoptions
options for     tcp{master}: 0x55b0885ec428
getoption       keepalive       nil     getsockopt failed
getoption       reuseaddr       nil     getsockopt failed
getoption       tcp-nodelay     nil     getsockopt failed
getoption       tcp-keepidle    nil     getsockopt failed
getoption       tcp-keepcnt     nil     getsockopt failed
getoption       tcp-keepintvl   nil     getsockopt failed
lua: ./test/tcp-getoptions:17: attempt to index a nil value
stack traceback:
        ./test/tcp-getoptions:17: in function 'options'
        ./test/tcp-getoptions:23: in main chunk
        [C]: ?
```

macOS:

```
$ ./test/tcp-getoptions
options for     tcp{master}: 0x7fd235808228
getoption       keepalive       nil     getsockopt failed
getoption       reuseaddr       nil     getsockopt failed
getoption       tcp-nodelay     nil     getsockopt failed
lua: ./test/tcp-getoptions:13: bad argument #1 to 'getoption' (unsupported option `tcp-keepidle')
stack traceback:
        [C]: in function 'getoption'
        ./test/tcp-getoptions:13: in function 'options'
        ./test/tcp-getoptions:23: in main chunk
        [C]: ?
```

ewestbrook:
[...]

PR #299, opened 2020-03-27
fix test/tcp-getoptions

ewestbrook:
In response to #297.  Work in progress.

Adding `socket =` to `require"socket"` fixes the reported error, but now I see others.

```
$ ./tcp-getoptions
options for     tcp{master}: 0x7f8a92804a28
getoption       keepalive       nil     getsockopt failed
getoption       reuseaddr       nil     getsockopt failed
getoption       tcp-nodelay     nil     getsockopt failed
lua: ./tcp-getoptions:13: bad argument #1 to 'getoption' (unsupported option `tcp-keepidle')
stack traceback:
        [C]: in function 'getoption'
        ./tcp-getoptions:13: in function 'options'
        ./tcp-getoptions:23: in main chunk
        [C]: ?
```

### DIFF
--- a/test/find-connect-limit
+++ b/test/find-connect-limit
@@ -10,7 +10,7 @@ ulimit -n
 You'll probably need to be root to do this.
 ]]
 
-require "socket"
+socket = require "socket"
 
 host = arg[1] or "google.com"
 port = arg[2] or 80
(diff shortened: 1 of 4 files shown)


jbbjarnason:
tcp-keepidle is only supported on unix socket if I remember it correctly, maybe that is the case ?

Issue #302, opened 2020-03-29
`socklen_t` already defined on mingw32-w32api-dev 5.3


DarkWiiPlayer:
While trying to install luasocket on a fairly new windows installation, I started getting the following error:

```
In file included from src/socket.h:18,
                 from src/inet.h:18,
                 from src/luasocket.c:30:
src/wsocket.h:14:13: error: conflicting types for 'socklen_t'
   14 | typedef int socklen_t;
      |             ^~~~~~~~~
In file included from src/wsocket.h:12,
                 from src/socket.h:18,
                 from src/inet.h:18,
                 from src/luasocket.c:30:
c:\mingw\include\ws2tcpip.h:399:22: note: previous declaration of 'socklen_t' was here
  399 | typedef unsigned int socklen_t;
      |                      ^~~~~~~~~
```

After digging through files for a few hours (honestly, having no idea what I was doing, as I don't know and don't care much about mingw or windows in general), I found that there was apparently an update that adds this type definition to the header file.

From what I can tell, the typedefs should be mostly compatible (`int` vs. `unsigned int`). Assuming the value can't be negative, which sounds reasonable, given the type name.

The commit that introduced the changes in the [git repo](https://osdn.net/projects/mingw/scm/git/mingw-org-wsl/summary):

```
commit a45cb9690d6298067a8cf9e5d69b9f006833034d
Author: Keith Marshall <keith@users.osdn.me>
Date:   Mon Jan 20 18:03:57 2020 +0000
```

The [LuaSec](https://github.com/brunoos/luasec) module seems to have the exact same problem.

DarkWiiPlayer:
As a quick fix, adding the following at the end of the rockspec works:

```lua
build.patches = {
['Fix mingw API change'] = [[
--- old/src/wsocket.h       2020-03-29 12:11:42.904053000 +0200
+++ new/src/wsocket.h       2020-03-29 12:10:06.829148800 +0200
@@ -11,7 +11,7 @@
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
-typedef int socklen_t;
+//typedef int socklen_t;
 typedef SOCKADDR_STORAGE t_sockaddr_storage;
 typedef SOCKET t_socket;
 typedef t_socket *p_socket;]]
}
```

DarkWiiPlayer:
I just dug through the mingw project again and found that the typedef isn't new, but it used to be just `int` in the past and now changed to `unsigned int`

[...]

PR #303, opened 2020-03-30
Remove obsolete typedef from compatibility file

DarkWiiPlayer:
MinGWs `ws2tcpip.h` already defines `socklen_t`, making this redundant.
What's more, in a recent version of MinGW this has changed to `unsigned int`, breaking compatibility. See issue #302 

I *tried* looking through the git log to see if there was a reason for this being there, but all I could find was in `f3305405` it was added without a comment, under the commit message

> Compiles and runs on linux and windows, using DLLs!

That was in 2003 using who knows which version of MinGW 😖

Thus, I assume it is OK to just remove this typedef altogether, but it'd be nice if someone could actually confirm this (Ideally, clarify why this was added in the first place)

### DIFF
--- a/src/wsocket.h
+++ b/src/wsocket.h
@@ -11,7 +11,6 @@
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
-typedef int socklen_t;
 typedef SOCKADDR_STORAGE t_sockaddr_storage;
 typedef SOCKET t_socket;
 typedef t_socket *p_socket;


ewestbrook:
Is there a macro we can test, for use in an #ifdef, in order to continue supporting older environments that might still require this typedef?

DarkWiiPlayer:
I actually have no idea. I went through the code and didn't find anything and I grepped for `version`, `53`, `5.3` and a few other combinations to see if there was some convenient typedef somewhere in the codebase but didn't find anything.

But again, I don't know if it's even necessary for the typedef to be there in the first place.

diegonehab:
Why have we never heard of this problem when compiling under Windows? Is this a mingw problem?

DarkWiiPlayer:
Well, for one because the MinGW version that changed it was released in the beginning of this month (Version 5.3, iirc). Other than that, no idea. if by "compiling under windows" you mean microsofts compiler, I have never used that, so I couldn't tell you either.

limjiregister:
How to resolve this issue?

alerque (changes requested):
This doesn't feel right. I'm not on Windows to test and poke around, but I don't see any logic for why this *shouldn't* be speced out here. Either it needs to be guarded behind some sort of check for whether it is already defined or something else needs to happen. Perhaps the correct way to resolve it is just let it be defined twice, but make it an unsigned int internally as well so we don't have conflicts?

PR #304, opened 2020-03-31
Add vcpkg installation instructions

PhoebeHui:
LuaSocket is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build LuaSocket, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for LuaSocket and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/doc/installation.html
+++ b/doc/installation.html
@@ -79,6 +79,21 @@ distribution directory structure:</p>
 would be replaced by <tt>core.so</tt>.  
 </p>
 
+<p>
+Alternatively, you can build and install LuaSocket using vcpkg dependency manager:
+
+<pre class=example>
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh
+./vcpkg integrate install
+./vcpkg install luasocket
+</pre>
+
+The luasocket port in vcpkg is kept up to date by microsoft team members and community contributors.
+If the version is out of date, please create an issue or pull request on the vcpkg repository.
+
+</p>
 <h3>Using LuaSocket</h3>
 
 <p> With the above setup, and an interpreter with shared library support,


ewestbrook:
I vote no.  There are already many ways to build and install LuaSocket.  Let's improve those, if needed, rather than advertise another.

avently:
@PhoebeHui hello. Looks like there is no way to build luasocket with lua 5.1 using your package manager, am I right? There is only the latest version of Lua available. How to work around this issue?

PhoebeHui:
@avently, if you want to use lua 5.1, you can follow below steps:

1. Open ./ports/lua/portfile.cmake, update the URLS and FILENAME with lua 5.1
2. ./vcpkg install lua, it would fail since the hash is not correct, you can copy the 'Actual Hash', and replace the hash value.
3. Commnent out PATCHES vs2015-impl-c99.patch since it's only for latest revision.
4. Check the src and headers in ports/lua/CMakeLists.txt file, remove these file doesn't in lua 5.1, and add the missing one. 
5. /vcpkg install lua

doc/installation.html
```diff
@@ -79,6 +79,21 @@ <h3>Directory structure</h3>
...
 </p>
 
+<p>
+Alternatively, you can build and install LuaSocket using vcpkg dependency manager:
+
+<pre class=example>
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh
+./vcpkg integrate install
```
alerque, line 89: I'm fine with mentioning this can be installed via `vcpkg`. It looks like their package is even still current. However I'm not fine with including documentation about how to install `vcpkg`. If people don't know that they should be going to the source for that. Lets just stick to a basic mention of `vcpkg install luasocket`.

alerque (changes requested):
Trim this down to a minimal mention of the package availability, not a tutorial on how to use vcpkg itself.

doc/installation.html
PhoebeHui: Sure, thanks for your review!

PR #305, opened 2020-04-14
Allow CC and LD to be overrriden

diegonehab:


ewestbrook:
At first I wondered whether we should do this similarly to MYCFLAGS et al, but then realized that while MYCFLAGS and such might need to be added to, CC and LD would only be replaced, if anything.  I like it.

diegonehab:
Turns out this is not necessary if the user calls make CC=bla, but I dind't know about that until just now. And since the initial value of CC is cc, I think this won't work as expected. :/ Maybe we should revert the change. Shall I do it?

ewestbrook:
Oh!  That's right, and I should have realized that too.  Yes, please.

diegonehab:
Done. :/

diegonehab:
Sorry about the noise.

PR #334, opened 2021-06-22
Fixes an issue with aux buffer init overwriting optional parameters in receive().

### LINKED ISSUE #331: HTTP calls do not work with lua 5.4.3
# Luasocket bug with Lua 5.4.3

## Short Description

HTTP calls do not work with Luasocket 5.4.3 and luasocket.

## System Info

Lua: 5.4.3
Luasocket: LuaSocket 3.0-rc1
Distro Archlinux

## Test code:

Requires `lua-socket` and `lua-sec` to be installed.
If testing using lua5.3, then `lua53-socket` `lua53-sec` need to be installed.

```lua
local http = require "socket.http"

print("---")
print("Testing https")

print(http.request("https://ffl.shoeb.pw"))

print("---")
print("Testing http")


print(http.request("http://ffl.shoeb.pw"))

print("---")
print("End http")
```

## Expected output:

```shell
$ lua5.3 t2.lua 
---
Testing https
Welcome to ffl.shoeb.pw. We shall be serving you soon!!!        200     table: 0x55a7c9e0f4d0   HTTP/1.1 200 OK
---
Testing http
Welcome to ffl.shoeb.pw. We shall be serving you soon!!!        200     table: 0x55a7c9e06510   HTTP/1.1 200 OK
---
End http
```
## Observed output:
```shell
$ lua t2.lua 
---
Testing https
Welcome to ffl.shoeb.pw. We shall be serving you soon!!!        200     table: 0x5639643d1800   HTTP/1.1 200 OK
---
Testing http
lua: /usr/share/lua/5.4/socket/http.lua:54: bad argument #1 to 'receive' (string expected, got light userdata)
stack traceback:
        [C]: in function 'socket.http.request'
        t2.lua:12: in main chunk
        [C]: in ?
```

## Explanation

HTTP call does not work. Diving into the source code to see what might be the cause yields the line:

```
 50 local function receiveheaders(sock, headers)
 51     local line, name, value, err
 52     headers = headers or {}
 53     -- get first line
 54     line, err = sock:receive()
 55     if err then return nil, err end
 56     -- headers go until a blank line is found
 57     while line ~= "" do
```

As you can see in the above code block, line 54 does not pass any arguments, so the argument that is passed will be `self` because of the `:` used in the function call.

I can confirm that this was working before my recent update to lua 5.4.3
So, mostly the recent update is what broke it as luasocket has not been updated for quite some time now.

Not sure if there needs to be some support for 5.4.3 that needs to be built into luasocket

pkulchenko:
This is the same issue that was fixed in MIME code in f329aae7.

Closes #331.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -103,11 +103,14 @@ int buffer_meth_send(lua_State *L, p_buffer buf) {
 * object:receive() interface
 \*-------------------------------------------------------------------------*/
 int buffer_meth_receive(lua_State *L, p_buffer buf) {
-    int err = IO_DONE, top = lua_gettop(L);
+    int err = IO_DONE, top;
     luaL_Buffer b;
     size_t size;
     const char *part = luaL_optlstring(L, 3, "", &size);
     timeout_markstart(buf->tm);
+    /* make sure we don't confuse buffer stuff with arguments */
+    lua_settop(L, 3);
+    top = lua_gettop(L);
     /* initialize buffer with optional extra prefix
      * (useful for concatenating previous partial results) */
     luaL_buffinit(L, &b);


danog:
+

monrax:
+1

src/buffer.c
```diff
@@ -103,11 +103,14 @@ int buffer_meth_send(lua_State *L, p_buffer buf) {
...
 int buffer_meth_receive(lua_State *L, p_buffer buf) {
-    int err = IO_DONE, top = lua_gettop(L);
+    int err = IO_DONE, top;
     luaL_Buffer b;
     size_t size;
     const char *part = luaL_optlstring(L, 3, "", &size);
     timeout_markstart(buf->tm);
+    /* make sure we don't confuse buffer stuff with arguments */
+    lua_settop(L, 3);
+    top = lua_gettop(L);
```
Zash, line 113: Isn't this line redundant and always equivalent to `top = 3` here? I'm not actually sure what the point of the `top` variable is, but this would change the behavior.

pkulchenko: The behavior in the patch is different from `top=3`, as `lua_settop()` will extend the stack to make sure that subsequent buffer allocations don't clobber optional parameters. The way this is done is consistent with the original patch from Diego in f329aae7 (which just missed this particular instance).

alerque (approved):
I've tested this across several systems (all Linux but a mix of Ubuntu & Arch Linux) and versions of Lua and this does appear to fix the bug it claims to fix. Given the limited breadth that I can test but the certain knowledge that this has completely blocked usage on current Lua (5.4.3 and 5.4.4) for Linux users I think it's fair to go ahead with a merge and release. If we must iterate past that then we must.

PR #343, opened 2021-10-11
Fix win64 compiler warnings on gai_strerror

rpatters1:
Issue:

Win64 produces compiler warnings because the Win64 version of `gai_strerror` returns `WCHAR *`. This is clearly a problem, because returning `WCHAR *` to a caller expecting `char *` will (usually) lead to the display of no more than one character.

Solution:

Implemented `GAI_STRERROR` macro in `socket.h` and replaced all calls to `gai_strerror` with the macro. For Win64 (and Win32) the macro explicitly calls `gai_strerrorA`.


alerque:
This sounds pretty reasonable, but I don't have access or expertise to test on Windows. What's the status on this? Any chance you'd be able to help add Windows to our CI matrix (PR #356 adds macOS but the action currently used has issues on Windows, perhaps an upstream fix there would help us implement tests here so we can verify this PR works and nothing else breaks.

alerque (approved):

PR #345, opened 2021-10-17
Fix argumentless receive call on Lua 5.4 #331

Zash:
It seems that on Lua 5.4 when calling conn:receive() without arguments,
a lightuserdata related to the luaL_Buffer ends up in position 2. Since
this is not a number, luaL_optstring() complains.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -113,7 +113,9 @@ int buffer_meth_receive(lua_State *L, p_buffer buf) {
     luaL_buffinit(L, &b);
     luaL_addlstring(&b, part, size);
     /* receive new patterns */
-    if (!lua_isnumber(L, 2)) {
+    if(top < 2){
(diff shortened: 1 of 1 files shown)


alerque:
If I'm not mistaken this is an alternative fix to #331 that is already covered by what I just merged from @pkulchenko in #334, is that correct?

alerque:
@zash If I've overlooked something please do pipe up. I went with Paul's patch because it seemed like logic more in keeping with how the original maintainer had dealt with similar issues and already had more feedback/review, but I'm happy to reevaluate if there is an argument to make for this being a better fix.

In any case thanks for contributing, and sorry it took so long to get attention.

Zash:
@alerque yes, alternative fix that I submitted in my haste before noticing that #334 was already there.

PR #315, opened 2020-08-06
ci: add Visual C++ and MinGW jobs

kou:
https://github.com/kou/luasocket/actions/runs/196950032 is a result of this.

MinGW job is failed because #312 isn't included. If we have #312, the job is passed: https://github.com/kou/luasocket/actions/runs/196944672

This uses Lua 5.4 for Visual C++ and Lua 5.1 for MinGW.

This has `CMakeLists.txt` to build Lua 5.4 with Visual C++. It's written me from scratch because there isn't the official Lua build system for Visual C++. There are some projects that writes `CMakeLists.txt` such as https://github.com/LuaDist/lua . I searches them when I'm writing this. It seems that the official Lua will not support CMake because there is no consensus to use CMake. I searched lua-users mailing list: http://lua-users.org/cgi-bin/namazu.cgi?query=cmake+visual&idxname=lua-l&max=20&result=normal&sort=date%3Alate

Should we use other `CMakeLists.txt` implementation rather than the `CMakeLIsts.txt` in this pull request? I think that the `CMakeLists.txt` in this pull request is simple enough to maintain.

`test/hello.lua` is only used for testing. We can add more tests as follow-up tasks.

PR #347, opened 2021-11-03
Update cmake build tools

jingyuexing:


alerque:
I'm curious why you call this an "update" since it seems to be adding CMake tooling from scratch. Did this originate somewhere else? I'm happy to consider this, but I'm a little hesitant to add an alternative build system before we clean up the one we have and get automated testing working properly. Also I'd like to hear what the use-case for this is that would make in worth the maintenance overhead above just having a normal LuaRocks based install. Lastly, will you be around to keep iterating on this as we possibly shuffle things around in the build. It would be good if we had somebody that was actually going to maintain the CMake side of things if we're going to include it here.

alerque (changes requested):
I'm not in a position to test this, so (in addition to answering the questions in my other comment) automated tests will be required before we even consider merging. Consider rebasing or a merge from master after testing related PRs like #356 have landed. We'll want some sort of check to that confirms the CMake system is functional with the matching current source tree.

jingyuexing:
> I'm curious why you call this an "update" since it seems to be adding CMake tooling from scratch. Did this originate somewhere else? I'm happy to consider this, but I'm a little hesitant to add an alternative build system before we clean up the one we have and get automated testing working properly. Also I'd like to hear what the use-case for this is that would make in worth the maintenance overhead above just having a normal LuaRocks based install. Lastly, will you be around to keep iterating on this as we possibly shuffle things around in the build. It would be good if we had somebody that was actually going to maintain the CMake side of things if we're going to include it here.

I believe there are many people who have the same problem as me, which is "cannot compile socket library on windows". Because makefiles are better for development environment on *nix systems

alerque:
#415 added some Windows CI jobs checking the build on the Windows side of things, at least with MSCV. #315 also has some cmake related stuff with a different route to build. @leso-kn Any input here? I'm not a Windows guy at all. I'm happy to facilitate contributions but I'm kind of out of my depth on the Windows side of things.

PR #351, opened 2022-02-01
src/ltn12.lua: remove duplicated codes

jyoui:


src/ltn12.lua
```diff
@@ -22,9 +24,6 @@ _M.source = source
 _M.sink = sink
 _M.pump = pump
 
-local unpack = unpack or table.unpack
```
alerque, line 25: If `unpack()` is being used anywhere in this file, this actually serves a purpose as a shim for different Lua version compatibility and should not be removed. If it isn't being used at all then of course it is cruft, but please confirm that because it isn't the same issue as the "duplicated" code removed in the rest of this commit.

alerque:
I'm willing to give this another review whet it is ready, but I think it might have overlapped some with the linting fixes merged in #354. Rebasing or merging from master should take care of that and also bring in the CI lint jobs to this PR.

src/ltn12.lua
Tieske: the exact same declaration is at line 12, so this is fine

Tieske (approved):

PR #355, opened 2022-03-19
Update URLs in (and to) SCM rockspec, start an epoch

alerque:
Closes #218
Closes #233
Closes #330

PR #356, opened 2022-03-19
Add CI workflows to test builds

alerque:
Follow up to linter job, not (yet) automated deploys this is just build checks

(Based on #355 so I don't have to redo rockspec location stuff, should not be merged first.)


alerque:
This is not parity with what the obsolete Travis CI config was doing — and beyond since it is test building across 3 more Lua versions and macos in addition to the limited selection of aging Luas on Linux in the Travis matrix.

@Tieske I still have this in draft mode because I want to sort out and merge that rockspec related PR before this one, but this is ready to go as soon as you take a gander at that.

alerque:
Also note this doesn't do any of the automated release stuff yet, just runs the regression test suite on various platforms.

daurnimator (commented):
> chore: Rename rockspec dir to be plural 

Why?

.github/workflows/build.yml
```diff
@@ -0,0 +1,61 @@
...
+          # lua testsupport.lua
+          # lua tftptest.lua
+          # lua udpconnectclnt.lua
+          # lua udpconnectsrvr.lua
+          # lua unixdgramclnt.lua
+          # lua unixdgramsrvr.lua
+          # lua unixstreamclnt.lua
+          # lua unixstreamsrvr.lua
+          # lua utestclnt.lua
+          # lua utestsrvr.lua
```
daurnimator, line 60: Why are these here and commented out?

alerque: Because I was playing around with files that were in tests but not part of the current suite. I'll remove them before taking this out of draft mode.

Tieske:
> > chore: Rename rockspec dir to be plural
> 
> Why?

consistency. @alerque did some great work on automating releases upon tagging, but this does require some consitency (at least that's what I assume is the reason)

Tieske (approved):

alerque:
Yes, what @Tieske said. It is a lot easier to reuse stuff like that when projects are consistent about stuff like that. And as long as I'm directly involved in the release process I'd rather normalize the project than add exceptions and variants to all the related tooling. Obviously both could be done, but given that grammatically a plural or collective noun is called for anyway fixing that seems like a better option all around.

PR #364, opened 2022-03-23
cleanup; move and remove obsolete files

Tieske:
This PR is very arbitrary, hence I'd like @diegonehab to review it before anything.

- deletes' a powerpoint file
- moves the logo file
- deletes the`./gem` folder (seems some latex related to render ltn docs??)
- merge `./etc` into `./samples`
- deletes `NEW`, since that info is already listed in the docs
- merges `FIX`, `TODO` and `WISH` into a single file `TODO.md` (latter one could use a thorough review probably)
- updates the LTNs from wiki to markdown format for proper rendering in Github

It is best reviewed on a commit by commit basis


alerque:
Most of this looks good, but can we wait to merge until after a release? I'm kind of trying to minimize changes so that the next tag is a safe-harbor, then we can overhaul a bit more aggressively.

I would suggest getting rid of the TODO stuff entirely and migrating the content to issues (perhaps starting with a single bulk checklist issue, then opening individual issues for anything that seems to merit them.

Tieske:
totally fine with that, but seeking input from @diegonehab as to what is ok to do, and what not. (after the safe-harbor tagging)

Tieske:
@diegonehab any chance you can have a look at this? it's mostly moving/renaming/deleting files. Should be really straightforward to review on a commit-by-commit basis.

alerque (approved):

Tieske:
@alerque this has been open for a while now. I think it's still ok, and should be good to merge. Any second thoughts? otherwise I'll merge it later.

alerque:
Nope no second thoughts here, LGTM.

Issue #324, opened 2021-01-10
Please ship `mbox.lua` in rock


rrthomas:
This useful module is shipped by Debian, for example, but not in the LuaRock. It seems to be the only Lua module in the sources that is not shipped.

Tieske:
(asking since it is not documented)

rrthomas:
It parses [mbox](https://en.wikipedia.org/wiki/Mbox) files.

alerque:
Pardon me if I'm being dense, but why is a networking / socket library a good home for a file format parser? That sounds like something that should have a different home. Parsing an over the while protocol is one thing, parsing an on disk file format seems to me to be a category error. What utility does Debian get out of it being in this package?

alerque:
Having looked this over I am still of the opinion that it should be broken out into a different rock ... but in the mean time given the way we are using it for tests and the example code that provides I think it's entirely reasonable to ship it with the rock until such a time as it can be reasonably installed as a dependency from Luarocks.

rrthomas:
Thanks for doing this!

PR #365, opened 2022-03-23
cleanup; move 'mbox.lua' to tests

Tieske:
mbox.lua is located in the main source directory, but only used from the smtptest.

@diegonehab does that `mbox.lua` file have any use beyond the test? any reason to include it in the distribution as a regular module?


alerque:
See #324.

Lets keep this PR unmerged until after a release just so downstream packagers don't have anything substantial to change to switch from the assorted non-tagged commits they are currently packing to a tagged commit, then start fixing this kind of stuff.

Tieske:
ah, that explains it, except for what it exactly does. Needs docs. Closing this here, continuing in #324 .

PR #370, opened 2022-03-26
Change 3.0-rc1 to 3.0.0

robert-scheck:
Usually bumping the version should be a pre-release activity rather than a post-release one. The additional copyright year update is to match commit 58f94e47d2f7a7cfc0aa2034a1b9ce679ce2731c.


src/luasocket.h
```diff
@@ -10,8 +10,8 @@
 /*-------------------------------------------------------------------------* \
 * Current socket library version
 \*-------------------------------------------------------------------------*/
-#define LUASOCKET_VERSION    "LuaSocket 3.0-rc1"
-#define LUASOCKET_COPYRIGHT  "Copyright (C) 1999-2013 Diego Nehab"
+#define LUASOCKET_VERSION    "LuaSocket 3.0.0"
+#define LUASOCKET_COPYRIGHT  "Copyright (C) 1999-2022 Diego Nehab"
```
daurnimator, line 14: The last commit from Diego is in 2019.
We should probably add something more like "luasocket contributors"?

robert-scheck: Okay, I guess it might be better to just address the version here and to move the copyright to a separate issue?

alerque: There are a number of other places the license and copyright need updating. Let me take care of than in a dedicated PR, lets just cleanup the version stuff here (still puzzled how I missed thise with all the grepping I did for references to the rc, but hey...)

alerque (changes requested):
Just the version issues in the commit please...

alerque (approved):

PR #374, opened 2022-04-01
Fix #373

### LINKED ISSUE #373: Sumneko's Lua language server shows a warning for using a lowercased global variable
[This ](https://github.com/lunarmodules/luasocket/blob/1d61853ab84f0724502205f45849c2347d6a49ac/src/ltn12.lua#L18)is the offending line.

image

Since it uses EmmyLua for annotations, it doesn't recognize `-- luacheck: ignore`.

Some context: Sumneko's Lua language server is the most widely used extension in VS Code for Lua support. It would be nice if we didn't get that warning.

C3pa:
Adds and EmmyLua annotation to skip diagnostics on the next line. Here is some more [info](https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations#diagnostic).

### DIFF
--- a/src/ltn12.lua
+++ b/src/ltn12.lua
@@ -15,6 +15,7 @@ local select = select
 
 local _M = {}
 if module then -- heuristic for exporting a global package table
+    ---@diagnostic disable-next-line:lowercase-global
     ltn12 = _M  -- luacheck: ignore
 end
 local filter,source,sink,pump = {},{},{},{}


alerque (changes requested):
Lets follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification for commit messages. Your PR and issue have relevant issue but the commit message will not be very informative in a changelog later. Lets try something like:

```
chore: Annotate lowercase global to avoid EmmyLua warning
```

alerque (approved):

Tieske:
I do not like this. LuaSocket should not be changed because downstream projects have (extremely) arbitrary linting rules. In this case for globals to be required in allcaps.

My 2cts.

C3pa:
Well, it's not a problem for me. I can easily apply a patch on top. I just offered this to see if there is some interest.

alerque:
@Tieske You do realize we already have a similar lint ignore rule here and a couple other places in our source to accommodate `luacheck`, right? The stunt we're pulling on that line of code is just a bit unconventional and understandable makes linter(s) unhappy. The *reason* EmmyLua* is unhappy is (IMO) sillier than the reason luacheck is unhappy and I would much rather see a config file in the project root to disable this silly rule for the whole project, but as far as I know that isn't supported and one rule to accommodate a lot of people and potential contributors that happen to be using VScode doesn't seem like too much of a burden to me.

Tieske:
It's a slippery slope imo. The luacheck exceptions are here because this project actively uses it to check against its own guidelines for proper code.

I agree it's not too much hassle, but it's just not a responsibility of this project. If anything then EmmyLua should exclude those files from their linter. And even more so because they use a different one from LuaCheck. Anyone casually looking at the LuaSocket code will be baffled by those lines.

Just my 2cts, don't want to block this solely on my opinion though.

C3pa:
Since there is no interest in this, I'll close this and apply the patch downstream.

PR #386, opened 2022-08-23
fix(http): use the right protocol for proxies

### LINKED ISSUE #385: Requesting HTTPS over an HTTP proxy doesn't work
When requesting an HTTPS url over an HTTP proxy, LuaSocket tries to initiate a TLS session with the proxy server.

This works with mitmproxy, but not with other proxies.

Is it intended behavior ?

Otherwise, I can submit a PR to use the HTTP create function when an HTTP proxy is used.

Max1Truc:
Fixes #385.

Needs some testing before merging.

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -219,9 +219,11 @@ local function adjustproxy(reqt)
     local proxy = reqt.proxy or _M.PROXY
     if proxy then
         proxy = url.parse(proxy)
-        return proxy.host, proxy.port or 3128
+        proxy.port = proxy.port or 3128
+        proxy.create = SCHEMES[proxy.scheme].create(reqt)
+        return proxy.host, proxy.port, proxy.create
     else
-        return reqt.host, reqt.port
+        return reqt.host, reqt.port, reqt.create
     end
 end
 
@@ -291,7 +293,10 @@ local function adjustrequest(reqt)
     end
 
     -- ajust host and port if there is a proxy
-    nreqt.host, nreqt.port = adjustproxy(nreqt)
+    local proxy_create
+    nreqt.host, nreqt.port, proxy_create = adjustproxy(nreqt)
+    if not reqt.create then nreqt.create = proxy_create end
+
     return nreqt
 end
 


Max1Truc:
After testing, this was too simple to be true.

At least one proxy program (Privoxy) requires the user to use the CONNECT verb.

I'm now going to use CONNECT on HTTPS proxies.

Max1Truc:
I could not figure a clean way of handling https over an https proxy, but it seems to be a luasec problem.

This PR thus only changes the behavior of http over an https proxy or https over an http proxy.
For http over http the behavior was already correct.
For https over https one would need to ssl.wrap over an existing tls (not tcp) socket.

Max1Truc:
By the way, this is a bit weird.
I realize this would be cleaner if I submit a PR to luasec enabling https.tcp to reuse a socket.

What do you think ? Would it be worth it ?

alerque:
> I realize this would be cleaner if I submitted a PR to luasec enabling https.tcp to reuse a socket.
> What do you think ? Would it be worth it ?

If that's the hold up to making this cleaner and it's correct behavior then yes I think it is worth a shot.

Would that by any chance be already covered in [this open PR](https://github.com/brunoos/luasec/pull/2)?

Max1Truc:
> If that's the hold up to making this cleaner and it's correct behavior then yes I think it is worth a shot.

Ok, I'll have a look.

> Would that by any chance be already covered in [this open PR](https://github.com/brunoos/luasec/pull/2)?

Sadly, it seems session reuse doesn't solve our problem.
Looking at a [sample file from the PR](https://raw.githubusercontent.com/brunoos/luasec/8f695ee2502b5860f62454b5c00bf8ff5c8d6606/samples/caching/client.lua), it looks like reusing the same TLS session across multiple TCP sockets.

alerque:
Where are we at with this?

Max1Truc:
I don't have the time for now, but it needs a Pull Request to luasec to allow wrapping tcp over tcp, and after my initial research such a PR would need to use C and OpenSSL.

This Pull Request is thus in pause, waiting for an update from luasec I guess.

alerque:
Okay. Let me know if there is something actionable. I don't have time to dig into luasec and contribute anything upstream at the moment, but will help facilitate this getting through when possible. Also even if it is a little messy given a "proper" fix won't be timely if there is a way to shim things on this end that isolates the temporary code in a non-breaking way but that can get people by while they wait pending proper upstream support in the future we could consider that too.

alerque:
> I don't have the time for now, but it needs a Pull Request to luasec to allow wrapping tcp over tcp,

The LuaSec project has recently been migrated to this org and we have a few more maintainers around to poke it and facilitate contributions if there is still something that needs to be fixed on that end of things.

Just FYI.

Max1Truc:
Thanks, I'll have a look.

Max1Truc:
This PR should be focused on using the right protocol while connecting to proxies, and the last force-push reflects this.

I'll handle the CONNECT verb in another PR.

I'll let you know when I'll have thoroughly tested this implementation.

Max1Truc:
Ok great, this PR looks good to me. Feel free to review it.

Previously, luasocket wouldn't work with anything tls and proxy, not luasocket supports http over an https proxy.

I tested using tinyproxy as a proxy, and socat to forward tls-encrypted packets back-and-forth between luasocket and tinyproxy.

I'm not how how to create automated tests for it though.

alerque (approved):

alerque:
This project has serious issues with testing and we need to work up a way to make Busted setup and tear down some server and client processes. A proxy would need to be part of that of course.

Max1Truc:
Is there an issue for adding proper tests to luasocket ?

alerque:
I thought there was but I couldn't find to link in my last comment. Feel free to open one!

PR #390, opened 2022-08-30
minor fix

zhaozg:


alerque:
Thanks for the contribution! I merged these commits manually so I could fix the commit messages so they show up in the changelog when I cut releases. Next time when you open a PR if you do it from a branch instead of your fork's `master` (or whatever the default) branch is GitHub will actually make it easier for maintainers to tweak things like this before accepting PRs. That way they will show up as accepted too. No big deal, just so you know going forward.

See 48164b5 for my merge. This should be in the next release, whenever that is. You can use the `scm-3` rockspec to start using it now.

zhaozg:
Got that, thank you.

siffiejoe:
Re "typo": `hstrerror` and `strerror` are two different functions.

alerque:
@siffiejoe True, but everything I saw suggested the singleton use of `hstrerror` in this code was an oversight. The exact same situations in the wsocket module uses `strerror` as does every other situation in the same usocket module where it might have come up. It looks to me like it would work either way but for consistency both alternative socket modules handling return messages the same way seemed to make sense. If I'm overlooking something I'd be happy to review.

siffiejoe:
From `man gethostbyname` on Linux:
```
...
RETURN VALUE
       The  gethostbyname()  and gethostbyaddr() functions return the hostent structure or a null pointer if an error occurs.  On error, the h_errno variable holds an error number.  When
       non-NULL, the return value may point at static data, see the notes below.

ERRORS
       The variable h_errno can have the following values:

       HOST_NOT_FOUND
              The specified host is unknown.

       NO_DATA
              The requested name is valid but does not have an IP address.  Another type of request to the name server for this domain may return an answer.  The constant NO_ADDRESS is a
              synonym for NO_DATA.

       NO_RECOVERY
              A nonrecoverable name server error occurred.

       TRY_AGAIN
              A temporary error occurred on an authoritative name server.  Try again later.
...
```

Test program for those four values which can be found in the `netdb.h` header file:
```c
#include <string.h>
#include <stdio.h>
#include <netdb.h>

static int error_numbers[] = {
  HOST_NOT_FOUND, TRY_AGAIN, NO_RECOVERY, NO_DATA
};

int main(void)
{
  size_t i = 0;
  for (i = 0; i < sizeof(error_numbers)/sizeof(error_numbers[0]); ++i)
  {
    printf("'%s'   vs.   '%s'\n", strerror(error_numbers[i]), hstrerror(error_numbers[i]));
  }
  return 0;
}
```

Results:
```
siffiejoe@Merkur:~ :) gcc errnos.c 
siffiejoe@Merkur:~ :) ./a.out 
'Operation not permitted'   vs.   'Unknown host'
'No such file or directory'   vs.   'Host name lookup failure'
'No such process'   vs.   'Unknown server error'
'Interrupted system call'   vs.   'No address associated with name'
siffiejoe@Merkur:~ :)
```

The function `socket_hoststrerror` is used together with `inet_gethost` in `inet.c`, and `inet_gethost` calls `socket_gethostbyname` or `socket_gethostbyaddr` in `usocket.c`, which in turn call `gethostbyname` or `gethostbyaddr`, respectively. So at least on Linux this change will lead to somewhat confusing error messages.

alerque:
Reverted.

PR #395, opened 2022-10-02
Fix relative redirection on https

### LINKED ISSUE #393: http: relative redirection fails for https scheme
Given than the [`Location`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) header can refer to relative url, the following code prevent relative https redirection

https://github.com/lunarmodules/luasocket/blob/26b524e1d7b9a00045882d3c0f25486485b1d6a8/src/http.lua#L298-L306

`scheme` is `nil` on relative url, I suggest to either use reqt as `default` parameter to `url.parse` (maybe side-effect to dig into first), or skip the downgrade test in this case, since the scheme is preserved by `tredirect` code anyway.

https://github.com/lunarmodules/luasocket/blob/26b524e1d7b9a00045882d3c0f25486485b1d6a8/src/http.lua#L325-L328

nheir:
Location header can now be relative: https://httpwg.org/specs/rfc9110.html#field.location

Fix #393

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -300,6 +300,8 @@ local function shouldredirect(reqt, code, headers)
     if not location then return false end
     location = string.gsub(location, "%s", "")
     if location == "" then return false end
+    -- the RFC says the redirect URL may be relative
+    location = url.absolute(reqt.url, location)
     local scheme = url.parse(location).scheme
     if scheme and (not SCHEMES[scheme]) then return false end
     -- avoid https downgrades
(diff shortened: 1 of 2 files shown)


alerque (approved):

alerque (approved):
Thanks for taking the time to research and make this contribution.

MikuAuahDark:
Wouldn't this break the tests since it requires LuaSec (LuaSocket doesn't support HTTPS)?

alerque:
It did not. If it *should have* broken a test which test exactly do you think should have failed?

MikuAuahDark:
Exactly this HTTP redirection test. Isn't LuaSocket lacked support for HTTPS?

nheir:
indeed the `httptest.lua` test will fail if `luasec` is not installed (raising on the assertion from `http.lua`)

current `httptest.lua` is also making additional hypotheses about having a `luasocket-cgi` runing on `localhost:3128` serving the `cgi` folder tho

both assume the test environment to be a bit specific and test socket and http code at the same time.
in order to provide a reliable test without dependence to a third party, `httptest.lua` could use a ad-hoc fake socket that match luasocket api 
see [cpython/Lib/test/test_httplib.py](https://github.com/python/cpython/blob/main/Lib/test/test_httplib.py) as example

PR #418, opened 2023-11-08
Fix empty path parsing

### LINKED ISSUE #417: url.parse() doesn't correctly handle URL with no path component.
Here we have a URL with a port component (`:8071`) and a query string component (`?token=xxx&type=http-bulk`) but with the path component left out:

```lua
local url = require "socket.url"

local parsed = url.parse("https://listener-eu.logz.io:8071?token=xxx&type=http-bulk")

print(require("inspect")(parsed))
```

Result:

```lua
{
  authority = "listener-eu.logz.io:8071?token=xxx&type=http-bulk",
  host = "listener-eu.logz.io",
  port = "8071?token=xxx&type=http-bulk",
  scheme = "https"
}
```

I feel like one of the following must be true (probably need to read through an RFC to know for sure):

* This is a valid URL, but it's not parsed correctly
* This is not a valid URL for $reasons (path is required when query string is used or... something), and it should be rejected with an error

---

ref: https://github.com/Kong/kong/issues/11931

alerque:
Fixes #417.

How does this fix look? Simplistic but *more valid* than the current parsing I think.

### DIFF
--- a/src/url.lua
+++ b/src/url.lua
@@ -152,7 +152,7 @@ function _M.parse(url, default)
     url = string.gsub(url, "^([%w][%w%+%-%.]*)%:",
         function(s) parsed.scheme = s; return "" end)
     -- get authority
-    url = string.gsub(url, "^//([^/%?]*)", function(n)
+    url = string.gsub(url, "^//([^/%?#]*)", function(n)
         parsed.authority = n
         return ""
     end)
--- a/src/url.lua
+++ b/src/url.lua
@@ -152,7 +152,7 @@ function _M.parse(url, default)
     url = string.gsub(url, "^([%w][%w%+%-%.]*)%:",
         function(s) parsed.scheme = s; return "" end)
     -- get authority
-    url = string.gsub(url, "^//([^/]*)", function(n)
+    url = string.gsub(url, "^//([^/%?]*)", function(n)
         parsed.authority = n
         return ""
     end)


catwell:
It does look more correct but I am unsure what to do with the path. If I am not mistaken here the end result is that the path will be nil because of [this](https://github.com/lunarmodules/luasocket/pull/418/files#diff-e9df027ada3350cefac0a112fc2fcd600c37f6abc1e7fbb3557b18c77983b534R175). Is this what we want? Alternatives may be either `"/"` or an empty string I guess.

alerque:
Yes, the path will be `nil`, but at least the query string will have the right value now and socket will be valid too. Given that path already returns nil in other cases if not set (i.e. empty per spec) that didn't seem like something I wanted to mess with. It might be more valid per spec to return an empty string, but that also might be a breaking change here! Ouch.

I didn't check if the fragment has the same issue, but I can do that.

alerque:
Yes, the same issue applied to fragments and I just fixed it the same way. I'm still hesitant to make path an empty string since I don't know what downstream usage might look like right now. Perhaps somebody is testing if a path exists by evaluating the path attribute for truthyness.

daurnimator (approved):

catwell:
@alerque I agree re. keeping the path `nil`.

I still think we should delete [the code parsing path parameters](https://github.com/lunarmodules/luasocket/pull/418/files#diff-e9df027ada3350cefac0a112fc2fcd600c37f6abc1e7fbb3557b18c77983b534R169-R173). This style of path parameters comes from [long deprecated RFC 1808](https://datatracker.ietf.org/doc/html/rfc1808#section-2.4.5) and is not used in modern HTTP, semicolons are a normal character (and here it may also cause issues with path parsing).

Tieske:
imho for all elements, path, fragment, etc, this test must hold;
```lua
assert(uri == url.build(url.parse(uri)))
```
Exceptions;
- path normalization
- escape/unescape differences

So for `path` being `nil` or `""`, both could work. But for `fragment` it should return a string, even an empty one, if `#` is present, as I don't think there is any other way to convey the fragment identifier `#` was present in the first place.

alerque:
@Tieske That's the way it works right now and we're not changing that, an empty string `""` is returned if the fragment marker is included even if no fragment is present.

@catwell I agree, but I think that should be in a separate PR as it is a much more involved fix & possibly breaking change. I'll get the party started with one after this merge and we can talk about the details there. It's part of the builder and other logic too.

PR #419, opened 2023-11-09
Remove deprecated URL params parsing

alerque:
See discussion begun in in #418

### DIFF
--- a/docs/url.html
+++ b/docs/url.html
@@ -60,7 +60,7 @@ An URL is defined by the following grammar:
 
 <blockquote>
 <tt>
-&lt;url&gt; ::= [&lt;scheme&gt;:][//&lt;authority&gt;][/&lt;path&gt;][;&lt;params&gt;][?&lt;query&gt;][#&lt;fragment&gt;]<br>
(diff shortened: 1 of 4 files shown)


alerque:
Is this what you had in mind @catwell?

Should this be considered a breaking change? That might affect what our next version number is.

Besides stopping to document this as a feature we should probably document the removal too.

catwell:
@alerque Yes, I had no idea it would be so impacting. It might be considered a breaking change but I would be surprised if people relied on this behavior. Tools typically implement the newer behavior e.g.:

```
$ trurl -g "{path"} "ftp://root:passwd@unsafe.org/pub/virus.exe;type=i"
/pub/virus.exe;type=i
```

catwell (approved):

alerque:
Also just for reference, the RFC that standardized params was 1995 and the RFC that obsoleted it with no mention of params was 2005.

I'm not *too* worried about dropping it without much ado, but we should still be careful. This library is used a lot of places and I want to keep the friction of updating to a minimum.

PR #420, opened 2023-11-13
Fix receiveheaders function

AMD-NICK:
It seems that error checking is needed here, otherwise in theory there will be an error in the while loop if `line` becomes false

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -62,7 +62,7 @@ local function receiveheaders(sock, headers)
         -- unfold any folded values
         while string.find(line, "^%s") do
             value = value .. line
-            line = sock:receive()
+            line, err = sock:receive()
             if err then return nil, err end
         end
         -- save pair in table


AMD-NICK:
Lua Test code:

```lua
local http = require("socket.http")
local _, err = http.request("http://dev:port", "")
print("err", err)
```

nodejs
```js
const net = require('net');

const server = net.createServer((socket) => {
    socket.write("HTTP/1.1 200 OK\r\n");
    socket.write("Content-Type: text/plain\r\n");
    socket.write("X-Custom-Header: key=val;\r\n");
	socket.write("    foo=bar;\r\n")
	socket.end(); // <<----------------------------------
	socket.write("    bar=baz;\r\n")
    socket.write("\r\n");
    socket.write("Hello, World!\r\n");
    
    socket.end();
});

const PORT = 8000;
server.listen(PORT, () => {
    console.log(`Server listening on port ${PORT}`);
});
```

Without fix:

`......./share/lua/5.1/socket/http.lua:50: bad argument #1 to 'find' (string expected, got nil)`

with:

`err     closed`

alerque (approved):

alerque:
Thanks for taking the time to track this down and contribute a fix.

PR #425, opened 2024-03-03
Add PSP as a supported platform

tpimh:


alerque:
Wow what an effort. I'm sure we can get this merged somehow, but the main catch is that I don't have anything PSP-ish to test on, so from upstream side we'd mostly be testing that nothing *else* broke and taking your word for how it goes on the PSP side.

It would be really helpful if anybody that was either running this on PSP or had ported anything else to the PSP and has experience with the platform could chime in here with a review. Do you happen to know any such folks you could send this way?

tpimh:
@alerque Yes, testing of PSP builds should be done on real hardware at the moment. That is until [PPSSPP](https://www.ppsspp.org/) networking code implements the required functions (other emulators exist, but I haven't looked into them).

There's a lot of stuff going on in this PR (initially I thought that would be easy). I should probably split it into multiple smaller PRs:

1. Optional IPv6 support
2. Static build support
3. PSP support

Some things in this PR really do not belong here: ~~sockaddr_storage~~, gethostname, getaddrinfo, herror... These parts should be added to [PSPSDK](https://github.com/pspdev/pspsdk/) and cleaned out of here.

For now, [luasocket is being built for PSP](https://pspdev.github.io/psp-packages/luasocket.html) only with Lua 5.1 and only statically. I want to have a build for every version of Lua available on PSP, and build dynamic .prx modules (Sony's equivalent of .so on Linux or .dll on Windows). I believe the required tooling is currently lacking, but it doesn't seem impossibly hard to implement the missing parts.

alerque:
Splitting up the PR into small chucks based on scope would be *great* and certainly allow some of this to land faster. May I suggest keeping this PR and branch around (probably switched to draft mode), and starting new branches for topic, then cherry picking relevant bits into the other branches for new PRs and review. Then you can keep playing with the whole system here and eventually keep rebasing this branch as individual parts land.

PR #430, opened 2024-04-27
Pass correct path length for abstract sockets

### LINKED ISSUE #216: No support for abstract namespace?
#I saw that this module supports unix socket and I've been trying to use it with abstract address. 

The below is the example code I tried:
address = '\0/var'
sock = socket.unix()
sock:connect(address)
which results in invalid argument.

Is there no support for abstract address?

Zash:
Since they start with NUL byte, strlen() does the wrong thing. The Lua API can already pass the real string length, so passing that on seems the sensible thing to do.

Fixes #216

### DIFF
--- a/src/unixdgram.c
+++ b/src/unixdgram.c
@@ -16,13 +16,6 @@
 
 #define UNIXDGRAM_DATAGRAMSIZE 8192
 
-/* provide a SUN_LEN macro if sys/un.h doesn't (e.g. Android) */
-#ifndef SUN_LEN
-#define SUN_LEN(ptr) \
-  ((size_t) (((struct sockaddr_un *) 0)->sun_path)  \
-   + strlen ((ptr)->sun_path))
-#endif
-
 /*=========================================================================*\
 * Internal function prototypes
 \*=========================================================================*/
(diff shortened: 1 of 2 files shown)


alerque (approved):

src/unixdgram.c
```diff
@@ -313,10 +313,9 @@ static int meth_getsockname(lua_State *L)
...
 * Turns a master unixdgram object into a client object.
 \*-------------------------------------------------------------------------*/
-static const char *unixdgram_tryconnect(p_unix un, const char *path)
+static const char *unixdgram_tryconnect(p_unix un, const char *path, size_t len)
 {
     struct sockaddr_un remote;
-    size_t len = strlen(path);
     if (len >= sizeof(remote.sun_path)) return "path too long";
     memset(&remote, 0, sizeof(remote));
     strcpy(remote.sun_path, path);
```
Zash, line 321: `strcpy()` and NUL-prefixed strings don't work, needs to be changed to something that takes a length. `strncpy()`? `memcpy()`?

Zash (commented):
> Pull request authors can’t request changes on their own pull request

Okay then!

Zash:
Sorry about that, now it actually works, not just _seems_ to work because `bind(sun_path="")` succeeds but does something else.

alerque (approved):

PR #434, opened 2024-06-28
chore: Remove the executable permission bit from some source files

uudiin:


alerque (commented):
Thanks for spotting this and taking the time to contribute. This looks quite reasonable for most of the files, but shouldn't the `*.cmd` and `*.bat` files still be marked as executable?

Tieske:
That flag has no relevance on Windows. So this should be good to go.

alerque (approved):

PR #438, opened 2024-09-11
`http.request`: possibility to defer body receiving if `sink` is not specified

### LINKED ISSUE #436: http.request api should be improved
Sinks (as well as filters) require response metadata to properly process data. There's a significant difference between receiving a 200 or 404 status code, or whether the content type is application/json, text/html, or an event stream.

Without this metadata, I’m forced to rely on unreliable heuristics in the sink/filter code, which is far from ideal.

It's unfortunate because the response metadata is already retrieved but simply isn't accessible within the sink.

johnd0e:
In this case, the first returned value is not `1`, but a function, taking two optional arguments: `sink` and `step`.
If the function is run without a `sink`, it simply closes the handle, ignoring the body.

Close #436

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -368,8 +368,16 @@ end
     local code, status = h:receivestatusline()
     -- if it is an HTTP/0.9 server, simply get the body and we are done
     if not code then
-        h:receive09body(status, nreqt.sink, nreqt.step)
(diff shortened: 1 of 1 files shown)


johnd0e:
@alerque Please review

PR #439, opened 2024-09-11
http.request: add missing handle close (HTTP/0.9 case)

johnd0e:
I am not 100% sure that this was an error, but for other cases handle is closed.

https://github.com/lunarmodules/luasocket/blob/1fad1626900a128be724cba9e9c19a6b2fe2bf6b/src/http.lua#L393


Tieske:
since http 0.9 is obsolete, I vote for not changing it. I'd expect it to do more harm than good.

PR #442, opened 2024-10-20
`socket.headers.canonic`: use dynamic canonization instead of limited set of predefined values

### LINKED ISSUE #440: Canonic header field capitalization does not work with all existing headers titles out of the box
The issue is that the list in [headers.lua](https://github.com/lunarmodules/luasocket/blob/master/src/headers.lua) is finite and cannot accommodate all possible header values. For instance, it does not include the commonly used header "Origin." I understand that keeping this list updated is challenging, especially given the existence of headers like "X-\*-\*."

Currently, the options are either to ignore this limitation (as most services handle lower-cased headers well) or to manage the headers manually.

My suggestion is to avoid storing headers altogether and instead normalize them on-the-fly. The existing `socket.headers` table could remain in place and serve as a cache.

johnd0e:
Fix #440


johnd0e:
Note: I would rather use "%f[^-]%l", but it doesn't match the beginning of the string.

To be even perfecter we should include other theoretically possible chars besides `%w`, but the RFC is not clear for me here https://www.rfc-editor.org/rfc/rfc7230#section-3.2

PR #445, opened 2024-12-12
Http header host for ipv6 address must be enclosed

Petr-kk:
in square brackets. Same as is in url.

url: http://[1080:0:0:0:8:800:200C:417A]/index.html
headers.host: [1080:0:0:0:8:800:200C:417A]

url.parse removes square brackets, I rather place them back here, than break something else depanding on url.parse


Tieske (changes requested):
tests?

PR #449, opened 2025-02-27
Add option for SO_EXCLUSIVEADDRUSE in Windows

Wires77:
Self-explanatory PR.  I found a use case for this in my project so I am submitting the PR upstream.  

See the MSDN docs for an explanation: https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse

### DIFF
--- a/docs/tcp.html
+++ b/docs/tcp.html
@@ -236,10 +236,20 @@ option names and values.
 <p class="parameters">
 <tt>Option</tt> is a string with the option name.</p>
 <ul>
+<li> '<tt>bindtodevice</tt>'</li>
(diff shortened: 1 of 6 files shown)


alerque:
This looks completely legit as an outsider, but I don't have access to any Windows environments to actually test this. Do we happen to have any other brave Windows users out there that could give this PR a test drive?

Tieske:
I rebased the branch. Ci is all green, so it builds on Windows. Personally confident that is good enough. Added a commit with the docs, including other options that were missing from the docs.

I'll leave the final verdict to @alerque .

alerque (approved):

Issue #451, opened 2025-03-02
UDP_DATAGRAMSIZE at compile time


sonoro1234:
Hello,

In order to change `UDP_DATAGRAMSIZE` at compile time but without modifying sources, it would be useful to have instead of:

https://github.com/lunarmodules/luasocket/blob/58c76080a0c654de906de9f554e2de45b1784929/src/udp.h#L20

```C
#ifndef UDP_DATAGRAMSIZE 
#define UDP_DATAGRAMSIZE 8192
#endif
```

alerque:
> Is there anything I can do to allow merging?

Getting on my notification radar when I wasn't traveling is what needed to happen. Sorry about the delay.

PR #452, opened 2025-03-02
let change UDP_DATAGRAMSIZE without touching source

sonoro1234:
as explained in issue #451

### DIFF
--- a/src/udp.h
+++ b/src/udp.h
@@ -17,7 +17,9 @@
 #include "timeout.h"
 #include "socket.h"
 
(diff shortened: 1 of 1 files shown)


Tieske (approved):
I like the change.

Just wondering if it might have to be placed in a generic configuration header file? But I wouldn't mind doing that as a follow up.

sonoro1234:
So, there is something missing to merge?

Tieske:
@alerque wdyt?

PR #455, opened 2025-05-27
add MSYS2/ucrt64 in makefile

Raffaello:
this will compile for MSYS2 ucrt64 if it might be helpful.

As at the moment using `luarocks` in MSYS2 ucrt64 doesn't work, this would help compiling luasocket from source.


alerque:
Thanks for taking the time to contribute!

PR #460, opened 2025-12-05
Add starttls support for SMTP

mbartlett21:


alerque (changes requested):
This looks alright as far as it goes, but it could use some documentation. It isn't clear at all from the documentation side how this would even be reached.

mbartlett21:
After looking at this some more I have decided to add it under a key under the main object and clear up how it works.

I have tested the current implementation and it is happy to deal with both Microsoft's and Google's SMTP receivers.

src/smtp.lua
```diff
@@ -260,12 +260,17 @@ end
...
 -- High level SMTP API
 -----------------------------------------------------------------------------
-_M.send = socket.protect(function(mailt, forcetls)
+_M.send = socket.protect(function(mailt)
     local s = _M.open(mailt.server, mailt.port, mailt.create)
     local ext = s:greet(mailt.domain)
-    if ext:find'STARTTLS' or forcetls then
-        -- we start tls now
-        ext = s:starttls(mailt.domain)
+    local tlsmode = mailt.tlsmode or "attempt"
```
alerque: Naming things is hard, but to me "attempt" implies that it might fall back to not using TLS, which is neither a satisfactory option nor what the code does.

Perhaps "none|auto|starttls|tls" would be a better set of options, then juggle the logic order a bit to first handle the if auto/starttls, elseif tls, else continue for none.

alerque (changes requested):

src/smtp.lua
mbartlett21: I've added a "tls" option now and renamed the "attempt" to "auto". For TLS I've split out a separate function now. I'm not sure about the wording for the docs.

src/smtp.lua
```diff
@@ -246,7 +268,19 @@ end
...
+    end
     local ext = s:greet(mailt.domain)
+    if tlsmode ~= "tls" and tlsmode ~= "none" then
+        if ext:find'STARTTLS' then
+            -- the greeting can change after security changes
+            ext = s:starttls(mailt.server, mailt.domain)
+        elseif tlsmode ~= "auto" then
+            s.try(nil, "server doesn't support STARTTLS")
+        end
+    end
```
alerque, line 291:
```suggestion
    local ext
    if tlsmode == "tls" then
        s:wraptls(mailt.server, mailt.domain)
        ext = s:greet(mailt.domain)
    else
        ext = s:greet(mailt.domain)
        if tlsmode ~= "none" then
            if ext:find("STARTTLS") then -- modes 'auto' and 'starttls' both try this
                -- the greeting can change after security changes
                ext = s:starttls(mailt.server, mailt.domain)
            elseif tlsmode == "starttls" then -- 'starttls' mode required but failed
                s.try(nil, "server doesn't support STARTTLS")
            end
        end
    end
```

Is this effectively the same logic? If so I think it's a bit easier to follow the logic with this conditional structure, no?

alerque (changes requested):

src/smtp.lua
mbartlett21: Applied it with you as the author. The logic is the same apart from invalid values are now treated as "auto" which seems ok.

Tieske: with tls, I think the only right way to handle it is to throw an error. Especially since "auto" might still fallback to plain text if the author makes a typo, eg. "TLS", instead of "tls", it might open an insecure connection.

mbartlett21: Are you happy with doing that inside an assert like so because it is a programming error or should it be with `s.try`?

```diff
+local valid_tlsmodes = { none = 1, auto = 1, tls = 1, starttls = 1 }
 
 ---------------------------------------------------------------------------
 -- High level SMTP API
 -----------------------------------------------------------------------------
 _M.send = socket.protect(function(mailt)
     local s = _M.open(mailt.server, mailt.port, mailt.create)
     local tlsmode = mailt.tlsmode or "auto"
+    assert(valid_tlsmodes[tlsmode], "invalid tlsmode")
```

---

Also, is there any sort of preferred process of squashing or rebasing etc for fixup commits?

Tieske: I'd make it a bit more friendly;

```lua
local valid_tlsmodes = { none = 1, auto = 1, tls = 1, starttls = 1 }

-- snip...

if not valid_tlsmodes[tlsmode] then
    error("Invalid tlsmode, expected one of 'tls', 'starttls', 'none', or 'auto'. Got '" .. tostring(tlsmode) .. "'")
end
```

alerque: > Also, is there any sort of preferred process of squashing or rebasing etc for fixup commits?

One way or another I want to end up with proper conventional commits, but I don't mind rewriting the branch myself if you want to push fixups here until everybody is happy with the implementation, then we can figure out what commits make sense.

Tieske (approved):
lgtm

Tieske (approved):

Issue #214, opened 2017-05-04
Build broken with LuaJIT


nzt4567:
Just to let you know, currently, both LuaJIT and luasocket define `luaL_setfuncs` which leads to link-time errors like 

```
/usr/local/lib/libluajit-5.1.a(lib_aux.o): In function `luaL_setfuncs':
/root/luajit-2.0/src/lib_aux.c:149: multiple definition of `luaL_setfuncs'
/usr/local/lib/lua/5.1/socket/libunix.a(compat.o):/root/luasocket/src/compat.c:7: first defined here
collect2: error: ld returned 1 exit status
```

Not sure if there is anything reasonable you can do about it though :(

pracj3am:
Maybe you should not define `LUA_VERSION_NUM=501` :)

diegonehab:
Did this solve the issue?

mpeterv:
I have successfully installed luasocket from master and using latest released version (3.0rc-1) both with LuaJIT 2.0.5 and LuaJIT 2.1.0-beta3 using LuaRocks without any errors.

siffiejoe:
Symbol lookup is different between static and dynamic linking. LuaJIT did add some symbols from Lua 5.2+, so with static linking you'll get in trouble if you also define those symbols for compatibility with older (PUC-Rio) Lua versions. Easiest workaround (if static linking is supposed to be supported) will probably be different names or a special prefix for the compatibility functions in LuaSocket, and maybe a conditional define so that you can still use the canonical names in the source.

diegonehab:
Which function are we talking about? luaL_setfuncs? If so, why not simply change the #if to delete the definition when a LuaJIT version that defines it is detected? If somebody sends me a pull request I'll be happy to merge it.

moteus:
I think problem in rockspec file.
On luarocks site scm rockspec does not include `compat.c` file
where this function defined for Lua 5.1/LuaJIT.

ewestbrook:
No errors here using current master with LuaJIT 5.0.2.  Is this still a problem?  Does anyone have solid reproduction steps or versions to test?

We're reviewing open items in preparation for a release.  If action is needed here, please add a comment.  Otherwise, this issue will be closed on or after 24-Feb-2019.

Thanks!

siffiejoe:
I'll look into it next weekend, but it doesn't seem anything has changed since then ...

ewestbrook:
[...]

PR #463, opened 2026-07-30
feat(receive): add maxsize argument to bound memory usage

Tieske:
`client:receive("*l")` and `receive("*a")` are unbounded: a peer that never sends a newline, or never closes, makes LuaSocket buffer until the process runs out of memory (e.g. `src/http.lua` reading response headers in a loop). Add an optional `maxsize` argument that caps the payload a single call may accumulate, including prefix.

- Hoist all argument validation ahead of `timeout_markstart()` so bad calls (`maxsize < 1`, `#prefix >= maxsize`, `numeric pattern > maxsize`) raise before any I/O and leave the socket untouched.
- `recvline`/`recvall` take a budget and return a new internal `BUF_OVERSIZED` code, surfaced to Lua as the "oversized" error alongside "timeout"/"closed", with the partial held in the third return value.
- `recvraw` is left untouched: argument checks make the cap unreachable for numeric patterns.
- Preserve three invariants: a timeout partial is always shorter than `maxsize` (safe to retry as `prefix`), completion beats the cap for `*a`, and no bytes are lost or skipped on overflow.
- `tcp.c`, `unixstream.c` and `serial.c` all share this code path unchanged.

Adds test coverage (argument errors, `*l`/`*a` boundaries, timeout/close at the cap, the drain idiom, numeric patterns, unix-stream mirror) and documents the new argument, error, and recovery idioms in `docs/tcp.html`.

### DIFF
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -134,7 +134,8 @@ int buffer_meth_receive(lua_State *L, p_buffer buf) {
     }
     if (!lua_isnoneornil(L, 4)) {
         double m = luaL_checknumber(L, 4);
-        luaL_argcheck(L, m >= 1, 4, "maxsize must be a positive number");
+        luaL_argcheck(L, m >= 1 && m < (lua_Number) ((size_t) -1), 4,
+            "maxsize must be a positive number");
         maxsize = (size_t) m;
         luaL_argcheck(L, size < maxsize, 4,
             "prefix length >= maxsize (drain with prefix=\"\" or raise maxsize)");
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -125,7 +125,8 @@ int buffer_meth_receive(lua_State *L, p_buffer buf) {
     /* ---- validation: must precede timeout_markstart() and any I/O ---- */
     if (numeric) {
         double n = lua_tonumber(L, 2);
-        luaL_argcheck(L, n >= 0, 2, "invalid receive pattern");
+        luaL_argcheck(L, n >= 0 && n < (lua_Number) ((size_t) -1), 2,
+            "invalid receive pattern");
         wanted = (size_t) n;
     } else {
         const char *p = luaL_optstring(L, 2, "*l");
--- a/src/buffer.c
+++ b/src/buffer.c
@@ -9,11 +9,23 @@
 * Internal function prototypes
 \*=========================================================================*/
 static int recvraw(p_buffer buf, size_t wanted, luaL_Buffer *b);
-static int recvline(p_buffer buf, luaL_Buffer *b);
-static int recvall(p_buffer buf, luaL_Buffer *b);
+static int recvline(p_buffer buf, luaL_Buffer *b, size_t budget);
+static int recvall(p_buffer buf, luaL_Buffer *b, size_t budget);
 static int buffer_get(p_buffer buf, const char **data, size_t *count);
 static void buffer_skip(p_buffer buf, size_t count);
 static int sendraw(p_buffer buf, const char *data, size_t count, size_t *sent);
+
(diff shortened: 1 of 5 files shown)


Tieske:
Just pinged a number of folks for a review. Some background; I just released a new version of [Copas](https://lunarmodules.github.io/copas/index.html#history) with a ton of security fixes. However some of the findings that triggered that originate from LuaSocket; unbounded reading on `"*l"` and `"*a"` patterns consuming all resources.

This PR is to fix that, such that I can continue fixing Copas. A follow up PR #467 fixes the same issues, but then here in LuaSocket.

In parallel there are some fixes ([#214](https://github.com/lunarmodules/luasec/pull/214), [#216](https://github.com/lunarmodules/luasec/pull/216)) in LuaSec, which @brunoos has been kindly reviewing, needed for the same purpose.

Hoping to get a quick review and a release. Thx!

src/buffer.c
```diff
@@ -105,8 +117,33 @@ int buffer_meth_send(lua_State *L, p_buffer buf) {
...
+        luaL_argcheck(L, n >= 0, 2, "invalid receive pattern");
+        wanted = (size_t) n;
+    } else {
+        const char *p = luaL_optstring(L, 2, "*l");
+        luaL_argcheck(L, p[0] == '*' && (p[1] == 'l' || p[1] == 'a'),
+                      2, "invalid receive pattern");
+    }
+    if (!lua_isnoneornil(L, 4)) {
+        double m = luaL_checknumber(L, 4);
+        luaL_argcheck(L, m >= 1, 4, "maxsize must be a positive number");
```
catwell: You could also add a check that maxsize does not overflow `size_t`.

catwell (approved):
LGTM

src/buffer.c
```diff
@@ -201,36 +242,61 @@ static int recvraw(p_buffer buf, size_t wanted, luaL_Buffer *b) {
...
+static int recvall(p_buffer buf, luaL_Buffer *b, size_t budget) {
     int err = IO_DONE;
     size_t total = 0;
     while (err == IO_DONE) {
         const char *data; size_t count;
         err = buffer_get(buf, &data, &count);
+        if (budget && count > budget - total) {   /* strictly more than fits */
+            count = budget - total;
+            luaL_addlstring(b, data, count);
+            buffer_skip(buf, count);
```
daurnimator, line 260: So this discards data when the budget is exceeded? That seems dangerous to then leave the socket open, as you don't know what data has been read+discarded. The only sensible solution here would be to either
a. keep the buffered data around for e.g. `:receive(2)` to see what was next or
b. close/cleanup the socket. or at least a half-shutdown.

src/buffer.c
Tieske: nice catch. Also fixed the same issue when parsing the pattern, a few lines up.

src/buffer.c
Tieske: No, the data read from the internal buffer is copied into `b` (a `luaL_Buffer`). When the error code `BUF_OVERSIZED` is returned, then the part copied into `b` is returned to the user as a partial result.

The user can retry by calling again, using the returned partial as `prefix` and a new (higher) budget.

Tieske: @daurnimator if you agree, then please resolve, to satisfy the merge rules. Thx!

daurnimator: I haven't verified but I trust you.

Tieske: added 2 tests to show the behaviour: https://github.com/lunarmodules/luasocket/pull/463/commits/4863f32b358a8d533f629084a86fb4932ebbd2f3

alerque (approved):

PR #466, opened 2026-08-03
Add HTTP response headers callback to allow for early abort or different sink depending on response

Frenzie:
Sample usage here: <https://github.com/koreader/koreader-base/pull/2485>.

### DIFF
--- a/src/http.lua
+++ b/src/http.lua
@@ -360,6 +360,7 @@ local trequest, tredirect
         headers = reqt.headers,
         proxy = reqt.proxy,
         maxredirects = reqt.maxredirects,
+        response_headers = reqt.response_headers,
         nredirects = (reqt.nredirects or 0) + 1,
         create = reqt.create
     }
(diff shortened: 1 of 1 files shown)


src/http.lua
```diff
@@ -403,6 +404,17 @@ end
         return tredirect(reqt, headers.location)
     end
     -- here we are finally done
+    -- provide an opportunity to abort or replace the sink based on the response headers
+    if nreqt.response_headers then
+        local abort, sink = nreqt.response_headers(code, headers, status)
```
Tieske, line 409: I think we should pass in another parameter; the raw socket. This can be used for websockets for example.

```suggestion
        local abort, sink = nreqt.response_headers(code, headers, status, h.c)
```

That said, I don't like the return values. Exactly opposite to idiomatic Lua; ok, err.
So instead of `local abort, sink = ...` it should be `local ok, sink = ...`, where `sink` is the error-object if `ok` is falsy.

then below in the error path, currently `if abort then ...`, we should pass back the error from the callback, not a success result as the code now does.

Tieske: addressed in #479

Frenzie: I'm not sure if I agree. It's not an error but merely a decision to stop. I could change it to something like:

```lua
local proceed, sink = nreqt.response_headers(code, headers, status, h.c)
if not proceed then
    h:close()
    return nil, code, headers, status
end
```

But then that distinction would be lost. The request didn't fail.

It does depend on the caller distinguishing between the two, but the caller is the one adding the callback in the first place.

Anyway, I'll think some more about it later.

Frenzie: > addressed in #479

Didn't see that. Alright. :-)

Frenzie:
Closing in favor of <https://github.com/lunarmodules/luasocket/pull/479>.

src/http.lua
Tieske: @Frenzie somehow missed your comments earlier. If you don't think this is the right solution, then please chime in on #479

PR #396, opened 2022-12-28
Support for Unix Sockets in Windows

MikuAuahDark:
This PR enables compilation of `socket.unix` module in Windows. Fixes #328.

Note that the Makefile change is untested, only the Visual Studio solutions. One thing I noticed that in Windows, `inet.c` must be compiled in Unix sockets. Rockspec changes is also untested.

Feel free to cherry-pick as appropriate if this PR as a whole lacked the quality and standards of the repository.

Issue #328, opened 2021-04-30
Enabling Unix Socket for Windows


MikuAuahDark:
Is it possible to enable codepath that allows Unix Socket to be created for Windows platforms? Recent Windows 10 supports Unix socket.

MikuAuahDark:
It's been more than a year and I don't hear any response, so might as well bump this one and share my findings.

So in one of our project, we recently updating to LuaSocket 3.1.0, so I took the opportunity to enable Unix sockets in Windows. It seems enabling this support is a matter of changing the included file. In Windows, `#include <afunix.h>` is used instead of `#include <sys/un.h>`. love2d/love@857cf406368975c2c88cba6d32dbaef0554aec4a.

Quick test of the Unix stream socket and it behaves as expected. Unfortunately Unix datagram is not supported.

Unix sockets in Windows has been supported since Windows 10 1809. Would there be an plan to enable unix sockets in Windows?

alerque:
We'd be happy to facilitate any PRs contributing to a better experience on the Windows side of things. Neither I nor any other current maintainer actively uses Windows though so researching, coding, and testing for this will need to come from community contributions. When you first posted this there wasn't an active maintainer, but I think we've made some progress on that front and there should be a path for improvements to integrated and released if they are contributed and tested.

PR #433, opened 2024-05-30
Added windows patch for mime.core to rockspec

goldenstein64:
Closes #429.

I made sure to modify the scm rockspec this time :)

I'm assuming this needs confirmation from another Windows user and testing on *nix platforms like before.

Issue #429, opened 2024-03-23
I have installed the lua54 version of window. And installed the luarocks of 3.11.0 version. Envirment path have been added. When I use the shell com


zack-incredible:
Installing https://luarocks.org/luasocket-3.1.0-1.src.rock

luasocket 3.1.0-1 depends on lua >= 5.1 (5.4-1 provided by VM: success)
x86_64-w64-mingw32-gcc -O2 -c -o src/mime.o -ID:\Programers_installation\Lua_Binary/include src/mime.c -DLUASOCKET_DEBUG -DWINVER=0x0501 -Ic:\windows\system32\include
x86_64-w64-mingw32-gcc -O2 -c -o src/compat.o -ID:\Programers_installation\Lua_Binary/include src/compat.c -DLUASOCKET_DEBUG -DWINVER=0x0501 -Ic:\windows\system32\include
x86_64-w64-mingw32-gcc  -shared -o C:\Users\zhaol\AppData\Local\Temp\luarocks_build-LuaSocket-3.1.0-1-2986800\mime\core.dll src/mime.o src/compat.o -Lc:\windows\system32 D:\Programers_installation\Lua_Binary\lua54.dll -lm
D:/Programers_installation/MingW64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/lib/../lib/dllcrt2.o:crtdll.c:(.text+0x16d): undefined reference to `_execute_onexit_table'
D:/Programers_installation/MingW64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/lib/../lib/dllcrt2.o:crtdll.c:(.text+0x8): undefined reference to `_initialize_onexit_table'
D:/Programers_installation/MingW64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/lib/../lib/dllcrt2.o:crtdll.c:(.text+0x38b): undefined reference to `_register_onexit_function'
D:/Programers_installation/MingW64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/lib/../lib/libmingw32.a(lib64_libmingw32_a-pseudo-reloc.o):pseudo-reloc.c:(.text.unlikely+0x28): undefined reference to `__acrt_iob_func'
[...]

PR #469, opened 2026-08-21
Build socket.unix on Windows

djungelorm:
Windows has carried AF_UNIX since Windows 10 1803 and Windows Server 2019, but `socket.unix` is still built only on unix, macOS and Haiku. This builds it on Windows too.

The module already reaches its sockets through the same `socket_*` layer as the TCP module, and `wsocket.c` implements that layer for Winsock, so the C changes are small:

 * `unixstream.c` and `unixdgram.c` take `sockaddr_un` from `<afunix.h>` rather than `<sys/un.h>`.
 * `bind`, `connect` and `sendto` are passed the whole address structure. Winsock reads the path as the null terminated string inside it rather than as far as the length covers.
 * `luaopen_socket_unix` starts Winsock. `require 'socket.unix'` can be the first thing a program does, and until now only `socket.core` called `socket_open()`.

The build system needs the module added in both places that decide what is built:

 * `luasocket-scm-3.rockspec` gains `modules["socket.unix"]` in the `win32`/`mingw32` branch, with `wsocket.c` in place of `usocket.c` and `ws2_32` to link against. `inet.c` is in the source list because `options.c` calls `inet_aton`, which on Windows is compiled into `inet.c` under `LUASOCKET_INET_ATON`; `inet.h` defines that macro itself under `_WIN32`, so nothing has to be passed on the command line. Without the file the module fails at link rather than at compile. This is the same dependency reported in #396.
 * `src/makefile` gains an `all-windows` target, which `win32`, `win64`, `mingw` and `ucrt64` invoke in place of `all`. It builds `$(UNIX_SO)` but not `$(SERIAL_SO)`: the serial module has no Windows implementation. `UNIX_OBJS` takes `$(SOCKET)` in place of `usocket.$(O)`, which is `usocket.$(O)` again on every unix platform, and gains `inet.$(O)` for the reason above. That object is redundant on unix, where `inet_aton` is in libc, but one object list is simpler than a conditional one for a file that costs nothing to link.

And there is a test, since #328 asked for evidence this works and no maintainer runs Windows:

 * `test/test_unix.lua` binds, listens, connects, accepts and exchanges a line over a stream socket, and sends a datagram between two dgram sockets everywhere except Windows, where it asserts instead that `socket.unix.dgram` reports the address family rather than returning a socket. It requires `socket.unix` on its own rather than after `socket`, so that the Winsock start up in `luaopen_socket_unix` is what is under test. The sockets are bound in the working directory: `sun_path` is 104 bytes on some platforms, and the temporary directory a Windows program is given is not one the test can work out from Lua.
 * `.github/workflows/build.yml` runs it alongside the other regression tests, so it runs on `ubuntu-22.04`, `macos-14` and `windows-2022` across the whole Lua matrix.
 * `.gitignore` gains `*.sock`, so a test that dies partway through leaves nothing behind.

Windows implements the stream sockets only. `socket.unix.stream` works; `socket.unix.dgram` compiles and returns `nil, "ai_family not supported"` in place of a socket, which is how the module reports an address family the platform will not open.

## Testing

The build workflow already runs on `windows-2022`, so this branch is built and tested on Windows by your own CI, on every Lua version in the matrix:

 * `luarocks make` builds `socket.unix` there now that the rockspec lists it, so the module is compiled and linked on Windows in this pull request rather than only on my machine. That step is what catches the `src/inet.c` dependency.
 * `test/test_unix.lua` then exercises it at run time, on Windows as well as on Linux and macOS.

Before that, the rockspec source list was checked by building it as a DLL with MSVC 2022 against Lua 5.1.5 on Windows 11, then loading it and opening, connecting, accepting and reading a unix domain socket. Dropping `src/inet.c` from that list fails with `unresolved external symbol inet_aton referenced in function opt_set_ip_multicast_if`. `test/test_unix.lua` was run against that build and passes.

kRPC (https://github.com/krpc/krpc) builds this patch into its lua client and its test suite runs on Windows in CI on every commit, connecting a client to a server over a unix domain socket, exchanging protocol buffer messages and closing the connection, alongside the same tests over TCP/IP. The same commit runs it all on Linux, where these changes are inert.

## Until this is released

kRPC needs `socket.unix` on Windows now, and a user who runs `luarocks install krpc` cannot wait for a release, so we intend to publish `luasocket-unix-windows` to luarocks.org: a rock that builds only this one module, only on Windows, from the v3.1.0 tarball with this change applied as an inline `build.patches` diff. It vendors nothing,
ships no other module, and exists only because no release carries this. We will stop publishing it once one does, and its issues point at the kRPC tracker rather than here, so it should not send you any traffic. Flagging it so it is not a surprise if you come across it, and happy to name it differently, or to drop it entirely, if you would rather it did not exist.

## References

 * #328, which asked for this.
 * #396, which is the earlier attempt at it. The C changes here are independent but reach the same conclusions, and the `inet.c` dependency was reported there first.

#396 also adds a `unix.vcxproj` and an entry in `luasocket.sln`, since one Visual Studio project cannot emit two DLLs. That is not here, because luarocks builds from the rockspec, but it is easy to add if you would rather the solution stayed complete.

### DIFF
--- a/src/unixdgram.c
+++ b/src/unixdgram.c
@@ -12,7 +12,11 @@
 #include <string.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <afunix.h>
+#else
 #include <sys/un.h>
+#endif
 
 #define UNIXDGRAM_DATAGRAMSIZE 8192
 
@@ -147,6 +151,11 @@ static int meth_sendto(lua_State *L)
     remote.sun_len = sizeof(remote.sun_family) + sizeof(remote.sun_len)
         + len + 1;
     err = socket_sendto(&un->sock, data, count, &sent, (SA *) &remote, remote.sun_len, tm);
+#elif defined(_WIN32)
+    /* Windows takes the whole address structure and reads the path as the null
+     * terminated string inside it, rather than as all the length covers. */
+    err = socket_sendto(&un->sock, data, count, &sent, (SA *) &remote,
+            sizeof(remote), tm);
 #else
     err = socket_sendto(&un->sock, data, count, &sent, (SA *) &remote,
 		   	sizeof(remote.sun_family) + len, tm);
@@ -268,6 +277,8 @@ static const char *unixdgram_trybind(p_unix un, const char *path, size_t len) {
         + len + 1;
     err = socket_bind(&un->sock, (SA *) &local, local.sun_len);
 
+#elif defined(_WIN32)
+    err = socket_bind(&un->sock, (SA *) &local, sizeof(local));
 #else
     err = socket_bind(&un->sock, (SA *) &local,
             sizeof(local.sun_family) + len);
@@ -323,6 +334,8 @@ static const char *unixdgram_tryconnect(p_unix un, const char *path, size_t len)
     remote.sun_len = sizeof(remote.sun_family) + sizeof(remote.sun_len)
         + len + 1;
     err = socket_connect(&un->sock, (SA *) &remote, remote.sun_len, &un->tm);
+#elif defined(_WIN32)
+    err = socket_connect(&un->sock, (SA *) &remote, sizeof(remote), &un->tm);
 #else
     err = socket_connect(&un->sock, (SA *) &remote,
             sizeof(remote.sun_family) + len, &un->tm);
--- a/src/unixdgram.c
+++ b/src/unixdgram.c
@@ -12,7 +12,7 @@
 #include <string.h>
 #include <stdlib.h>
 
-#ifdef _WIN32
+#if HAVE_WINDOWS_AFUNIX
 #include <afunix.h>
 #else
 #include <sys/un.h>
@@ -151,7 +151,7 @@ static int meth_sendto(lua_State *L)
     remote.sun_len = sizeof(remote.sun_family) + sizeof(remote.sun_len)
         + len + 1;
     err = socket_sendto(&un->sock, data, count, &sent, (SA *) &remote, remote.sun_len, tm);
-#elif defined(_WIN32)
+#elif HAVE_WINDOWS_AFUNIX
     /* Windows takes the whole address structure and reads the path as the null
      * terminated string inside it, rather than as all the length covers. */
     err = socket_sendto(&un->sock, data, count, &sent, (SA *) &remote,
@@ -277,7 +277,7 @@ static const char *unixdgram_trybind(p_unix un, const char *path, size_t len) {
         + len + 1;
     err = socket_bind(&un->sock, (SA *) &local, local.sun_len);
 
-#elif defined(_WIN32)
+#elif HAVE_WINDOWS_AFUNIX
     err = socket_bind(&un->sock, (SA *) &local, sizeof(local));
 #else
     err = socket_bind(&un->sock, (SA *) &local,
@@ -334,7 +334,7 @@ static const char *unixdgram_tryconnect(p_unix un, const char *path, size_t len)
     remote.sun_len = sizeof(remote.sun_family) + sizeof(remote.sun_len)
         + len + 1;
     err = socket_connect(&un->sock, (SA *) &remote, remote.sun_len, &un->tm);
-#elif defined(_WIN32)
+#elif HAVE_WINDOWS_AFUNIX
     err = socket_connect(&un->sock, (SA *) &remote, sizeof(remote), &un->tm);
 #else
     err = socket_connect(&un->sock, (SA *) &remote,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,7 @@ jobs:
           lua excepttest.lua
           lua test_bind.lua
           lua test_getaddrinfo.lua
+          lua test_unix.lua
           lua ltn12test.lua
           lua mimetest.lua
           lua urltest.lua
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,5 @@ Release
 *.swp
 *.suo
 x64
+*.sock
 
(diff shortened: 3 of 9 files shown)


src/unixdgram.c
```diff
@@ -12,7 +12,11 @@
 #include <string.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
```
daurnimator: Rather than just _WIN32; have a define `HAVE_WINDOWS_AFUNIX`
And have it default with something like:
```
#define HAVE_WINDOWS_AFUNIX (_WIN32 && some_version_check)
```


I guess that version check would use `_WIN32_WINNT`?

djungelorm: Good idea. I've added that. It first checks if the header exists on Windows (if the required macro exists), and failing that checks the Windows version number.

djungelorm:
Sorry took me a while to find the time to look at this.

I've had a look at both related PRs. Short summary: #433 isn't related at all, and #396 is a genuine duplicate that I think this PR can replace (now that I've folded in the one piece it had that this one was missing).

## #433

It adds `modules["mime.core"].libdirs = {}` for issue #429, where mingw builds of `mime.core` fail to link: LuaRocks passes `-Lc:\windows\system32`, so `ld` resolves against the system DLLs instead of mingw's import libraries. Nothing to do with AF_UNIX.

@goldenstein64 closed it themselves on 3 September, after working through it with @luau-project and concluding the fix belongs in LuaRocks' `external_deps_dirs` ordering - the mingw sysroot needs to come before `c:\windows\system32` — rather than in our rockspec. So there's nothing to merge or sequence there, and #429 stays open pending the LuaRocks side.

It does touch this PR at exactly one point. The third commit here applies the same `libdirs = {}` workaround to the new `socket.unix` module, for the same root cause - without it, mingw fails on an undefined `gai_strerrorA`. That mirrors what `socket.core` has done since #383. It isn't new debt, just the existing workaround applied consistently. Worth flagging that the #433 discussion floated *reverting* #383; if that's ever taken up, all of those `libdirs = {}` lines should come out together rather than one at a time.

## #396

This is the real overlap - same feature, same issue (#328). We independently reached the same core conclusions: build `socket.unix` on Windows from `wsocket.c` plus `inet.c` (`options.c` calls `inet_aton`, which Winsock doesn't provide, so the link fails without it), link `ws2_32`, and accept that Windows gives stream sockets only. Credit to @MikuAuahDark for getting there first, back in 2022. My bad for not noticing this before opening this PR...

However, there are differences, and reasons why I think we should take this PR over #396. Rather than just reading the diff, I built #396 locally and ran it, so the first and third points below are measured rather than assumed:

- **Winsock startup.** This is the one real defect, and it's the main reason to prefer this PR. `socket.unix` is a bare C module with no Lua wrapper, so `require "socket.unix"` never pulls in `socket.core`, and under #396 nothing calls `WSAStartup`. The `require` itself succeeds - the failure is deferred to the first socket creation, where `unix.stream()` returns `nil, "Successful WSAStartup not yet performed"`. That deferral is presumably why it went unnoticed: the module works fine whenever something else in the process required `socket` first, which is how it was originally tested. This PR calls `socket_open()` in `luaopen_socket_unix`, and `test_unix.lua` deliberately requires `socket.unix` on its own so that path stays covered - run against a #396 build, that test fails on its first `unix.stream()`.
- **`sockaddr_un`.** #396 hand-rolls the struct in a new `unixdef.h` and defines `AF_UNIX` itself, which was reasonable in 2022; this PR includes the SDK's `<afunix.h>` behind a `HAVE_WINDOWS_AFUNIX` probe, so we don't carry our own copy of a system ABI definition.
- **Address length.** Winsock reads `sun_path` as the null-terminated string inside the structure rather than as far as the length covers, so this PR passes the whole structure to bind/connect/sendto, where #396 keeps the Unix length computation. I want to be clear that this is a preference and not a fix: I tested #396's version and it works. With the `socket_open()` call added and nothing else changed, #396 passes the whole of `test_unix.lua` - bind, listen, connect, accept and data transfer included. I'd still rather pass the whole structure because it matches what the platform documents.
- **Tests.** This PR adds `test/test_unix.lua` and runs it in `build.yml`, so the module is exercised on Linux, macOS and Windows across 5.1–5.4, LuaJIT and OpenResty. #396 predates that Windows matrix and has no test.

They conflict rather than compose - both rewrite the same rockspec block and the same includes in `unixstream.c` and `unixdgram.c` - so it's one or the other.

## What I've changed here since you last reviewed

The one thing #396 had that this PR lacked was the MSVC project files, so I've cherry-picked them across (`unix.vcxproj` and the solution entry), with @MikuAuahDark noted as co-author in the commit message. I left out that commit's makefile changes: they fold the module into the `all` target for every platform, where this branch already reaches it through a separate `all-windows`.

I also added a CI job that builds `luasocket.sln`, because otherwise we'd be shipping a project file as unverified as it was in #396 - the workflow reaches everything through LuaRocks and never opens the solution, so `socket.vcxproj` and `mime.vcxproj` have never been built in CI either. The job compiles rather than tests, since the regression tests already run
against the LuaRocks-built modules; what's in question is only whether the solution still describes them.

I also merged in the latest changes from master branch, which brought in the new `test_unixdgram_receive_zero.lua`, which is skipped on Windows. That guard was added because `socket.unix` didn't build on Windows at all. After this PR the module loads there but has no datagram support, so the guard is still needed, just for a different reason.

## Suggested path

Merge this one and close #396 with credit to @MikuAuahDark. They have the same work, but this PR has the Winsock initialization and the test coverage. #433 is essentially unrelated and can be reviewed/merged separately.

MikuAuahDark:
If this PR is superior than mine (not gonna lie, my PR was almost 4 years old), then I am voting on merging this one and closing mine.

PR #478, opened 2026-08-31
Add an SSE client

Tieske:
Builds on top of #466 as an example use.

Note that the tests are very flaky.


catwell:
I have tested that branch against several APIs, it seems to work well!

The only problem as noted in the tests is that several real-world SSE servers do not use chunked encoding (for instance Firebase or nchan). There are several ways to deal with this, some inside luasocket and some at client-level, but it would be interesting to have an example of how it is expected to work.

